PR #3, opened 2017-06-01
HTTPMethod/HTTPResponseStatus enums need to be open

helje5:
I'd prefer to use Int constants for those, but we need to have an .other case at the minimum.

Whether the HTTPResponseStatus.other reasonPhrase should be optional is TBD. I'd say no, because the protocol actually requires that (though it has limited real-world use).

### DIFF
--- a/Sources/SwiftServerHttp/HTTPRequest.swift
+++ b/Sources/SwiftServerHttp/HTTPRequest.swift
@@ -35,9 +35,9 @@ public enum HTTPBodyChunk {
 }
 
 /// HTTP Methods handled by http_parser.[ch] supports
-public enum HTTPMethod: String {
-    // case custom(method: String)
-    case UNKNOWN
+public enum HTTPMethod : RawRepresentable {
+    /* be future-proof, http_parser can be upgraded */
+    case other(String)
     
     /* everything that http_parser.[ch] supports */
     case DELETE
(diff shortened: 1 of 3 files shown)


helje5:
We should probably drop RawRepresentable and move the API in the extension into the main enum. Why have that in an extension? If we *do* keep RawRepresentable, keeping 'code' seems superfluous.

Sources/SwiftServerHttp/HTTPRequest.swift
```diff
@@ -73,4 +73,82 @@ public enum HTTPMethod: String {
...
+            case "SUBSCRIBE":   self = .SUBSCRIBE
+            case "UNSUBSCRIBE": self = .UNSUBSCRIBE
+            case "PATCH":       self = .PATCH
+            case "PURGE":       self = .PURGE
+            case "MKCALENDAR":  self = .MKCALENDAR
+            case "LINK":        self = .LINK
+            case "UNLINK":      self = .UNLINK
+            default:            self = .other(rawValue)
+        }
+    }
```
michael-yuji, line 114: @helje5 I'm thinking since HTTPMethod are pretty well-defined, maybe we can use StaticString instead of String here??

helje5: I guess that may make some sense, but I'd rather have this as a separate PR given that it affects what `.other` can hold. E.g. a proxy may want to transfer arbitrary methods.

seabaylea:
If we're using enums for these then we definitely need to build in extensibility. I think we need to have a discussion about whether we switch these a struct + extension approach, but if there's no "-1"s I'll merge this as it improves what't there (and it can still be replaced if we change overall approach).

seabaylea:
Merging.

PR #4, opened 2017-06-02
Reduce unnecessary duplication on HTTPHeader by storing Index reference to original header data.

michael-yuji:


seabaylea:
LGTM. Any other comments?

helje5:
I don't care too much, but isn't that an unnecessary complication? The values backing the strings are already shared w/ that CoW stuff, right? There is no 'unnecessary duplication' in the previous version.

michael-yuji:
@helje5 That's true, I didn't think of that. The actual storage behind the buffer were never copied until mutation happen. But Int is still a one word struct vs String a 3 word struct. So it comes down to if we want to optimize storage overhead or computation overhead.

helje5:
It doesn't some down to that. It comes down to whether you want to complicate the code for small efficiency gains.

In case efficiency is the goal, I would still recommend to not parse the full header into Swift crap, but only do so on demand. I.e. the storage would be a byte buffer, the C parser would generate pointers into that, and Strings and such super high level objects would only be created if actually requested.

While I'm all in for doing any shenanigans in the implementation, the gains of this particular PR are very little IMO. If you enter Swift String/Dict/Array, all efficiency is essentially lost in the first place and such micro optimisations are just weird. Rather have Swift deal with making String/Dict/Array better if you play at that level (i.e. CoW on strings).

helje5:
P.S.: Oh, and my opinion is still that header keys and maybe even values should be an enum too (not an actual enum but Int consts to provide extensibility, but well). I like the idea of parsing values of known header (e.g. why store content-length as a String, that is neither, efficient nor type safe).

seabaylea:
Closing based on the discussion - we can always revisit in the future.

PR #5, opened 2017-06-07
Make StreamingParser.keepAliveUntil accessable outside the SwiftServerHttp module

shmuelk:
`StreamingParser.keepAliveUntil` is an `internal` variable and thus only accessible within the SwiftServerHttp module.

One can't create classes outside of SwiftServerHttp in the style of BlueSocketConnectionListener as a result.

This PR makes `StreamingParser.keepAliveUntil` `public`


seabaylea:
LGTM. This is required to switch to other networking layers (i.e. the goal to remove the BlueSocket dependency.)

PR #6, opened 2017-06-13
Convert HTTPVersion to a struct

seabaylea:
As per the proposal from @paulofaria, convert HTTPVersion from `(Int, Int)` to a struct. 

This allows the addition of Hashable, Equatable and CustomStringConvertible protocols - which haven't yet been added but can be if we're happy to move to a struct.

### DIFF
--- a/API.md
+++ b/API.md
@@ -64,7 +64,14 @@ public struct HTTPHeaders {
 }
 
 /// Version number of the HTTP Protocol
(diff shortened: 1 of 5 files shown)


tanner0101 (approved):
Great change. Type-aliases can't conform to protocols or be extended.

seabaylea:
Merging based on approval from @tanner0101 and discussion on the mailing list.

PR #12, opened 2017-06-19
Convert HTTPResponseStatus to a struct

aleksey-mashanov:
* `HTTPResponseStatus` is a struct now. It is better then an enum
  in this case because it could be possible to provide an alternative
  reason phrase as allowed by RFC7231:
  "The reason phrases listed here are only recommendations -- they
  can be replaced by local equivalents without affecting the protocol."

* `HTTPResponseStatus` struct is smaller then enum. Struct's size(stride)
  are 32(32) whereas enum's are 33(40).

* Removed conformance to protocol `RawRepresentable` because custom
  reason phrases conflict with the message of `RawRepresentable`:
  "With a RawRepresentable type, you can switch back and forth between a
  custom type and an associated RawValue type without losing the value
  of the original RawRepresentable type."

* Added conformance to `CustomStringConvertble` to provide an easy
  conversion to HTTP status line.

* Added conformance to `ExpressibleByIntegerLiteral` to provide an
  easy way to pass numeric value instead of `.humanReadable`.
  For many server-side developers 200, 404 and 500 have much more
  meaning then `.ok`, `.notFound` and `.internalServerError`.

* Added `class` property in conformance with RFC7231 recommendation:
  "HTTP clients are not required to understand the meaning of all
  registered status codes, though such understanding is obviously
  desirable.  However, a client MUST understand the class of any
  status code, as indicated by the first digit, and treat an
  unrecognized status code as being equivalent to the x00 status
  code of that class..."

* Replace `from` static method with initializer because in Swift it is
  more conventional to convert values using initializers.

### DIFF
--- a/Sources/HTTP/HTTPResponse.swift
+++ b/Sources/HTTP/HTTPResponse.swift
@@ -50,19 +50,19 @@ public enum HTTPTransferEncoding {
 
 /// Response status (200 ok, 404 not found, etc)
 public struct HTTPResponseStatus: Equatable, CustomStringConvertible, ExpressibleByIntegerLiteral {
-    public let code: UInt16
+    public let code: Int
     public let reasonPhrase: String
 
-    public init(_ code: UInt16, _ reasonPhrase: String) {
+    public init(_ code: Int, _ reasonPhrase: String) {
         self.code = code
         self.reasonPhrase = reasonPhrase
     }
 
-    public init(_ code: UInt16) {
+    public init(_ code: Int) {
         self.init(code, HTTPResponseStatus.defaultReasonPhrase(code))
     }
 
-    public init(integerLiteral: UInt16) {
+    public init(integerLiteral: Int) {
         self.init(integerLiteral)
     }
     
@@ -126,7 +126,7 @@ public struct HTTPResponseStatus: Equatable, CustomStringConvertible, Expressibl
     public static let notExtended = HTTPResponseStatus(510)
     public static let networkAuthenticationRequired = HTTPResponseStatus(511)
 
-    static func defaultReasonPhrase(_ code: UInt16) -> String {
+    static func defaultReasonPhrase(_ code: Int) -> String {
         switch code {
             case 100: return "Continue"
             case 101: return "Switching Protocols"
@@ -189,7 +189,7 @@ public struct HTTPResponseStatus: Equatable, CustomStringConvertible, Expressibl
         case serverError
         case invalidStatus
 
-        init(_ code: UInt16) {
+        init(_ code: Int) {
             switch code {
                 case 100..<200: self = .informational
                 case 200..<300: self = .successful
--- a/Sources/HTTP/HTTPResponse.swift
+++ b/Sources/HTTP/HTTPResponse.swift
@@ -49,224 +49,159 @@ public enum HTTPTransferEncoding {
 }
 
 /// Response status (200 ok, 404 not found, etc)
-public enum HTTPResponseStatus: RawRepresentable, Equatable {
-    /* be future-proof, new status codes can appear */
-    case other(statusCode: UInt16, reasonPhrase: String)
+public struct HTTPResponseStatus: Equatable, CustomStringConvertible, ExpressibleByIntegerLiteral {
+    public let code: UInt16
+    public let reasonPhrase: String
+
+    public init(_ code: UInt16, _ reasonPhrase: String) {
+        self.code = code
+        self.reasonPhrase = reasonPhrase
+    }
+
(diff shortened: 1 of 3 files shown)


helje5:
I prefer this over the enum as it is open w/o weird (different) syntax when the protocol changes. I also like the `class`.

Sources/HTTP/HTTPResponse.swift
```diff
@@ -49,224 +49,159 @@ public enum HTTPTransferEncoding {
...
-            case 504: self = .gatewayTimeout
-            case 505: self = .httpVersionNotSupported
-            case 506: self = .variantAlsoNegotiates
-            case 507: self = .insufficientStorage
-            case 508: self = .loopDetected
-            case 510: self = .notExtended
-            case 511: self = .networkAuthenticationRequired
-            default:  self = .other(statusCode: rawValue, reasonPhrase: "http_\(rawValue)")
-        }
+    public init(_ code: UInt16, _ reasonPhrase: String) {
```
tanner0101: Argument labels should be included here to compensate for weak type information.

tanner0101 (approved):
LGTM. I also think the `.class` property will be very useful.

Sources/HTTP/HTTPResponse.swift
aleksey-mashanov: I have thought about this a little. The main reason why no argument labels are provided is HTTP status line. It has exactly the same parameters and ordering. `"200 OK"` translates to `HTTPResponseStatus(200, "OK")` very intuitively. With labels it will become more Swifty but less HTTP.

seabaylea: I think this is another area where there's no right answer. @tanner0101 are you ok for me to merge this, and see what users think of the API usage when we have people try to use it?

seabaylea:
@aleksey-mashanov can you also update the API.md file as part of the PR?

Sources/HTTP/HTTPResponse.swift
tanner0101: screen shot 2017-06-20 at 2 29 27 pm

I'm just quoting Apple's Swift API design guidelines. 

@seabaylea yes, I think it's a great change either way!

aleksey-mashanov: * `HTTPResponseStatus(200, "OK")`
* `HTTPResponseStatus(200, reasonPhrase: "OK")`
* `HTTPResponseStatus(code: 200, reasonPhrase: "OK")`

Which one is better? I have no answer. The first one is strightforward. In the third label `code` has no more meaning then `value` - status is its status-code in the first place. And I don't know that to do with `init(_ code: UInt32)`. Should it look `HTTPResponseStatus(code: 200)` in this case?

tanner0101: Out of those three, I think `HTTPResponseStatus(code: 200, reasonPhrase: "OK")` adheres closest to the [Swift API design guidelines](https://swift.org/documentation/api-design-guidelines/). Both by compensating for weak type info as well as providing additional clarity over brevity.

> Clarity is more important than brevity. Although Swift code can be compact, it is a non-goal to enable the smallest possible code with the fewest characters.

In Vapor at least, it's [`HTTP.Status(statusCode: Int, reasonPhrase: String)`](https://github.com/vapor/engine/blob/d959474237a5bfe928239fe7b2cd6f62119badf5/Sources/HTTP/Status/Status.swift#L160)

However I don't think `HTTPResponseStatus(200, "OK")` is terribly confusing. Just doesn't feel Swifty. Which might not be enough to merit changing it. 😛

aleksey-mashanov: `Status(statusCode:` - is a "much of muchness".
OK, I have nothing against `HTTPResponseStatus(code: 200, reasonPhrase: "OK")`. Let's follow guidelines.

The vast majority of developers would omit reasonPhrase. How initializer with one argument should look like?
* `HTTPResponseStatus(200)` - to match conversion between types (imagine `Int("10")`)
* `HTTPResponseStatus(code: 200)`

aleksey-mashanov:
One more thing. I want to change status-code type from `UInt16` to `Int`. `Int` is better match Swift guidelines and does not enlarge `HTTPResponseStruct` size because of alignment.

Also done.

Sources/HTTP/HTTPResponse.swift
aleksey-mashanov: Done. `HTTPResponseStatus(code: 200, reasonPhrase: "OK")` and `HTTPResponseStatus(code: 200)`

seabaylea:
LGTM. Merging.

PR #13, opened 2017-06-20
Change HTTPRequest.target type form String to URL

huebs:
This PR changes the [`HTTPRequest.target`](https://github.com/swift-server/http/blob/develop/Sources/HTTP/HTTPRequest.swift#L15) property's type from `String` to `URL`. This accomplishes a few things:

  * Clarifies the type of data stored in the property.
  * Ensures the value is a valid URL.
  * Provides easy access to individual URL components.

### DIFF
--- a/Sources/HTTP/HTTPRequest.swift
+++ b/Sources/HTTP/HTTPRequest.swift
@@ -12,7 +12,7 @@ import Dispatch
 /// HTTP Request NOT INCLUDING THE BODY. This allows for streaming
 public struct HTTPRequest {
     public var method: HTTPMethod
-    public var target: String /* e.g. "/foo/bar?buz=qux" */
+    public var target: URL
     public var httpVersion: HTTPVersion
     public var headers: HTTPHeaders
 }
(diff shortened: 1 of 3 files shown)


helje5:
The problem is that the HTTP request line contains a Request-URI, not a URL. [RFC 2616](https://tools.ietf.org/html/rfc2616#section-5.1.2):

    Request-URI    = "*" | absoluteURI | abs_path | authority

E.g.

    OPTIONS * HTTP/1.1
    CONNECT host:port HTTP/1.1

Plus the overhead of using `URL` is higher than just a String. Though a String is not really correct either, in reality it should be a Data (percent escaping, user-agent dependent charsets etc.).

carlbrown:
We could make it a struct or an enum or something structured so that it's not "stringly-typed", but a URL is the wrong type for this.

seabaylea:
Hi @huebs. Are you ok if we shut this one down?

huebs:
Sure.

seabaylea:
Thanks @huebs. This was the right kind of idea, just unfortunately not applicable in this specific case.

PR #14, opened 2017-06-21
Make HTTPHeaders more user-friendly

aleksey-mashanov:
* Header name is a struct conforming to `ExpressibleByStringLiteral`.
  Standard header names are available through static constants.
  This allows to save some CPU on lowercasing and hash value
  calculation.

* Original headers are stored until first modification. Iterator
  of `HTTPHeaders` iterates over original headers while they are actual,
  in other case it iterates over underlying mutable dictionary.

* Default subscript returns values of all headers with the same
  name concatenated with comma [RFC7230, section 3.2.2], except
  for "Set-Cookie" for which only first value is returned.

* As-is values of all headers with same name can be accessed
  through subscript with label `valuesFor`.

* append() method is modified to accept dictionary literal.
  replace() method is added.

### DIFF
--- a/API.md
+++ b/API.md
@@ -52,15 +52,211 @@ public enum HTTPBodyChunk {
 }
 
 /// Headers structure.
-public struct HTTPHeaders {
-    var storage: [String:[String]]     /* lower cased keys */
-    var original: [(String, String)]   /* original casing */
-    let description: String
-    
-    public subscript(key: String) -> [String]
-    func makeIterator() -> IndexingIterator<Array<(String, String)>>
-    
-    public init(_ headers: [(String, String)] = [])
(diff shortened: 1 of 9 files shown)


seabaylea:
Hi @aleksey-mashanov in general this looks good to me. For stylistic reasons, could you separate out the protocol conformance of HTTPHeaders into extensions as we have for HTTPVersion?

Any views from anyone else?

aleksey-mashanov:
Done for top-level type `HTTPHeaders` only.

There are some problems with this stylistic guidance:
* It cannot be used if one of the requirements is stored.
* It cannot be applied to non-final classes.
* It works good only for top-level types. I tried to apply this rule to `HTTPHeaders.Name`. The result was very ugly and not readable.

seabaylea:
Thanks @aleksey-mashanov. Merging.

PR #15, opened 2017-06-25
Add Comparable conformance to HTTPVersion

seabaylea:
Add Comparable support for HTTPVersion plus associated tests.

### DIFF
--- a/Sources/HTTP/HTTPVersion.swift
+++ b/Sources/HTTP/HTTPVersion.swift
@@ -21,13 +21,12 @@ public struct HTTPVersion {
 }
 
 extension HTTPVersion : Hashable {
-
     public var hashValue: Int {
-        return major ^ minor
+        return (major << 8) | minor
     }
 
     public static func == (lhs: HTTPVersion, rhs: HTTPVersion) -> Bool {
-        return lhs.hashValue == rhs.hashValue
+        return lhs.major == rhs.major && lhs.minor == rhs.minor
     }
     
     public static func != (lhs: HTTPVersion, rhs: HTTPVersion) -> Bool {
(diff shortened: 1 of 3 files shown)


Sources/HTTP/HTTPVersion.swift
```diff
@@ -29,12 +29,44 @@ extension HTTPVersion : Hashable {
...
+            return lhs.minor < rhs.minor
+        }
+    }
+    
+    public static func <= (lhs: HTTPVersion, rhs: HTTPVersion) -> Bool {
+        return lhs < rhs || lhs == rhs
+    }
+
+}
+
```
aleksey-mashanov: Implementing `<` is enough for conformance to `Comparable`. Other operators are provided by stdlib in extension to `Comparable`. I don't see any performance advantages in implementing all operators by hand.

Sources/HTTP/HTTPVersion.swift
```diff
@@ -29,12 +29,44 @@ extension HTTPVersion : Hashable {
     public static func == (lhs: HTTPVersion, rhs: HTTPVersion) -> Bool {
         return lhs.hashValue == rhs.hashValue
     }
-
```
aleksey-mashanov: Hash values by design are guaranteed to be equal for two equal values but not vice versa.
So it is better to replace implementation of `==` with `lhs.major == rhs.major && lhs.minor == rhs.minor`.

For example, in current implementation hash values for versions 1.1 and 2.2 are equal. I suggest to implement `hashValue` in a more collision-resistant way, something like: `(major << 8) | minor`

michael-yuji: exactly.

Sources/HTTP/HTTPVersion.swift
seabaylea: Agreed - looking at the implementation in the stdlib:
https://github.com/apple/swift/blob/767e8f31ac8ab8186102dc95fe5916bb2d29bbb5/stdlib/public/core/Comparable.swift
there's no advantage to implementing these by hand. I'll update to remove.

Sources/HTTP/HTTPVersion.swift
seabaylea: I'll update to remove the potential hash collision and replace the implementation on `==`

Sources/HTTP/HTTPVersion.swift
```diff
@@ -21,20 +21,34 @@ public struct HTTPVersion {
...
+        return (major << 8) | minor
     }
 
     public static func == (lhs: HTTPVersion, rhs: HTTPVersion) -> Bool {
-        return lhs.hashValue == rhs.hashValue
+        return lhs.major == rhs.major && lhs.minor == rhs.minor
     }
-
+    
+    public static func != (lhs: HTTPVersion, rhs: HTTPVersion) -> Bool {
```
aleksey-mashanov: `!=` should be removed too.

seabaylea: Done.

seabaylea:
If there's no further review comments, I'll look to merge this tomorrow.

seabaylea:
Merging as there's no more review comments.

PR #16, opened 2017-06-27
Grab the lock when calculating the count property

ymanton:
Calculating the number of active connection listeners requires iterating through the underlying storage array, hence we need to take the lock the lock since new listeners may be appended to it concurrently. This fixes crashes that are seen under non-trivial load.

### DIFF
--- a/Sources/BlueSocketHTTP/BlueSocketSimpleServer.swift
+++ b/Sources/BlueSocketHTTP/BlueSocketSimpleServer.swift
@@ -163,7 +163,9 @@ class ConnectionListenerCollection {
     
     /// Used when shutting down the server to close all connections
     func closeAll() {
+        lock.wait()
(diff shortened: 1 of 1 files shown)


seabaylea:
Hi @ymanton. Whilst it may not be an issue due to timing - depending on the flow through the code -  shouldn't we also be protecting the access to `storage` in `closeAll()` as well?

ymanton:
You're right. It occurred to me at the time and I convinced myself that it wasn't necessary if one uses `closeAll()` sensibly.

In hindsight however if you're using it at shutdown the lock has little performance impact and if you're using it some other way you'll probably take contention over crashes so the module might as well protect itself. Change made.

seabaylea:
Thanks @ymanton. LGTM. Merging.

PR #18, opened 2017-06-30
.discardBody should call done()

seabaylea:
Running the following code to return 200 OK currently results in a hang:
```swift
    func serve(req: HTTPRequest, res: HTTPResponseWriter ) -> HTTPBodyProcessing {
        //Assume the router gave us the right request - at least for now
        res.writeResponse(HTTPResponse(httpVersion: req.httpVersion,
                                       status: .ok,
                                       transferEncoding: .chunked,
                                       headers: ["X-foo": "bar"]))
        return .discardBody
```
This is because `.discardBody` doesn't result in an eventual cal to `done()` to return the response.

I've added a test for this, along with a simple fix.


carlbrown (approved):
LGTM

PR #19, opened 2017-07-01
Simplification of HTTPResponse/HTTPResponseWriter

aleksey-mashanov:
* `HTTPResponse` is redundent. Its `status` and `headers` properties
  are moved to parameters of `writeHeader` method of `HTTPResponseWriter`
  instance. `transferEncoding` handling is moved to `HTTPHeaders` - it
  is not as simple as an enum and depends on used HTTP protocol version
  and capabilites set by a client in its request headers.
  `httpVersion` should not be configured by an outer scope - a server should
  use the HTTP version which was requested by a client.

* `writeResponse` and `writeContinue` are replaced by `writeHeader`.

* `writeTrailer` accepts trailers as `HTTPHeaders` struct.

* All `write*` methods accept a completion handler.

* `writeBody` accepts `UnsafeHTTPResponseBody` instead of just `Data`
  and `Dispatch`. It should be used only in the scope of a call and a
  copy should be made or value should be checked for `HTTPResponseBody`
  conformance before storing it anythere. This approch allows to reduce
  memory copying.

* `Result<POSIXError, Void>` enum is replaced with `Result`. In the
  first case `POSIXError` and `Void` are generic parameters names, not
  types. I think this behaviour was unintended.

Some problems with sync/async API still exist (and in the reference implementation of `HTTPStreamingParser` too). They should be addressed in an ongoing PR, after an agreement on the proposed changes.

### DIFF
--- a/API.md
+++ b/API.md
@@ -7,31 +7,44 @@ public struct HTTPRequest {
     public var headers : HTTPHeaders
 }
 
-/// HTTP Response NOT INCLUDING THE BODY
-public struct HTTPResponse {
-    public var httpVersion : HTTPVersion
-    public var status: HTTPResponseStatus
-    public var transferEncoding: HTTPTransferEncoding
-    public var headers: HTTPHeaders
-}
-
 /// Object that code writes the response and response body to. 
 public protocol HTTPResponseWriter : class {
-    func writeContinue(headers: HTTPHeaders?) /* to send an HTTP `100 Continue` */
-    
-    func writeResponse(_ response: HTTPResponse)
-    
-    func writeTrailer(key: String, value: String)
-    
-    func writeBody(data: DispatchData, completion: @escaping (Result<POSIXError, ()>) -> Void)
-    func writeBody(data: DispatchData) /* convenience */
-
-    func writeBody(data: Data, completion: @escaping (Result<POSIXError, ()>) -> Void)
-    func writeBody(data: Data) /* convenience */
-
-    func done() /* convenience */
-    func done(completion: @escaping (Result<POSIXError, ()>) -> Void)
+    func writeHeader(status: HTTPResponseStatus, headers: HTTPHeaders, completion: @escaping (Result) -> Void)
+    func writeTrailer(_ trailers: HTTPHeaders, completion: @escaping (Result) -> Void)
+    func writeBody(_ data: UnsafeHTTPResponseBody, completion: @escaping (Result) -> Void)
+    func done(completion: @escaping (Result) -> Void)
     func abort()
+}
+
+/// Convenience methods for HTTP response writer.
+extension HTTPResponseWriter {
+    public func writeHeader(status: HTTPResponseStatus, headers: HTTPHeaders)
+    public func writeHeader(status: HTTPResponseStatus)
+    public func writeTrailer(_ trailers: HTTPHeaders)
+    public func writeBody(_ data: UnsafeHTTPResponseBody)
+    public func done()
+}
+
+public protocol UnsafeHTTPResponseBody {
+    func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
+}
+
+extension UnsafeRawBufferPointer : UnsafeHTTPResponseBody {
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
+}
+
+public protocol HTTPResponseBody : UnsafeHTTPResponseBody {}
+
+extension Data : HTTPResponseBody {
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
+}
+
+extension DispatchData : HTTPResponseBody {
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
+}
+
+extension String : HTTPResponseBody {
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
 }
 
 /// Method that takes a chunk of request body and is expected to write to the ResponseWriter
--- a/Sources/HTTP/HTTPCommon.swift
+++ b/Sources/HTTP/HTTPCommon.swift
@@ -17,18 +17,7 @@ public protocol WebAppContaining: class {
     func serve(req: HTTPRequest, res: HTTPResponseWriter ) -> HTTPBodyProcessing
 }
 
-public enum Result<POSIXError, Void> {
-    case success(())
-    case failure(POSIXError)
-    
-    // MARK: Constructors
-    /// Constructs a success wrapping a `closure`.
-    public init(completion: ()) {
-        self = .success(completion)
-    }
-    
-    /// Constructs a failure wrapping an `POSIXError`.
-    public init(error: POSIXError) {
-        self = .failure(error)
-    }
+public enum Result {
+    case ok
+    case error(Error)
 }
--- a/Sources/HTTP/HTTPResponse.swift
+++ b/Sources/HTTP/HTTPResponse.swift
@@ -9,41 +9,27 @@
 import Foundation
 import Dispatch
 
-/// HTTP Response NOT INCLUDING THE BODY
-public struct HTTPResponse {
-    public var httpVersion: HTTPVersion
-    public var status: HTTPResponseStatus
-    public var transferEncoding: HTTPTransferEncoding
-    public var headers: HTTPHeaders
-    
-    public init (httpVersion: HTTPVersion, status: HTTPResponseStatus, transferEncoding: HTTPTransferEncoding, headers: HTTPHeaders) {
-        self.httpVersion = httpVersion
-        self.status = status
-        self.transferEncoding = transferEncoding
-        self.headers = headers
-    }
-}
-
 /// Object that code writes the response and response body to. 
 public protocol HTTPResponseWriter : class {
-    func writeContinue(headers: HTTPHeaders?) /* to send an HTTP `100 Continue` */
-    func writeResponse(_ response: HTTPResponse)
-    func writeTrailer(key: String, value: String)
-    func writeBody(data: DispatchData, completion: @escaping (Result<POSIXError, ()>) -> Void)
-    func writeBody(data: Data, completion: @escaping (Result<POSIXError, ()>) -> Void)
-    func done(completion: @escaping (Result<POSIXError, ()>) -> Void)
+    func writeHeader(status: HTTPResponseStatus, headers: HTTPHeaders, completion: @escaping (Result) -> Void)
+    func writeTrailer(_ trailers: HTTPHeaders, completion: @escaping (Result) -> Void)
+    func writeBody(_ data: UnsafeHTTPResponseBody, completion: @escaping (Result) -> Void)
+    func done(completion: @escaping (Result) -> Void)
     func abort()
 }
 
 /// Convenience methods for HTTP response writer.
 extension HTTPResponseWriter {
-    /// A convenience method for writing the supplied
-    /// `DispatchData` to the body of the HTTP response without
-    /// needing to supply a completion closure.
-    ///
-    /// - see: writeBody(data:completion:)
-    public func writeBody(data: DispatchData) {
-        return writeBody(data: data) { _ in }
+    public func writeHeader(status: HTTPResponseStatus, headers: HTTPHeaders) {
+        writeHeader(status: status, headers: headers) { _ in }
+    }
+
+    public func writeHeader(status: HTTPResponseStatus) {
+        writeHeader(status: status, headers: [:])
+    }
+
+    public func writeTrailer(_ trailers: HTTPHeaders) {
+        writeTrailer(trailers) { _ in }
     }
 
     /// A convenience method for writing the supplied
@@ -51,8 +37,8 @@ extension HTTPResponseWriter {
     /// needing to supply a completion closure.
     ///
     /// - see: writeBody(data:completion:)
-    public func writeBody(data: Data) {
-        return writeBody(data: data) { _ in }
+    public func writeBody(_ data: UnsafeHTTPResponseBody) {
+        return writeBody(data) { _ in }
     }
 
     /// A convenience method for signalling that the
@@ -63,11 +49,6 @@ extension HTTPResponseWriter {
     public func done() {
         done { _ in }
     }
-}
-
-public enum HTTPTransferEncoding {
-    case identity(contentLength: UInt)
-    case chunked
 }
 
 /// Response status (200 ok, 404 not found, etc)
@@ -223,7 +204,57 @@ public struct HTTPResponseStatus: Equatable, CustomStringConvertible, Expressibl
         }
     }
 
+    // [RFC2616, section 4.4]
+    var bodyAllowed: Bool {
+        switch code {
+            case 100..<200: return false
+            case 204: return false
+            case 304: return false
+            default: return true
+        }
+    }
+
+    var suppressedHeaders: [HTTPHeaders.Name] {
+        if self == .notModified {
+            return ["Content-Type", "Content-Length", "Transfer-Encoding"]
+        } else if !bodyAllowed {
+            return ["Content-Length", "Transfer-Encoding"]
+        } else {
+            return []
+        }
+    }
+
     public static func ==(lhs: HTTPResponseStatus, rhs: HTTPResponseStatus) -> Bool {
         return lhs.code == rhs.code
+    }
+}
+
+public protocol UnsafeHTTPResponseBody {
+    func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
+}
+
+extension UnsafeRawBufferPointer : UnsafeHTTPResponseBody {
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
+        return try body(self)
+    }
+}
+
+public protocol HTTPResponseBody : UnsafeHTTPResponseBody {}
+
+extension Data : HTTPResponseBody {
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
+        return try withUnsafeBytes { try body(UnsafeRawBufferPointer(start: $0, count: count)) }
+    }
+}
+
+extension DispatchData : HTTPResponseBody {
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
+        return try withUnsafeBytes { try body(UnsafeRawBufferPointer(start: $0, count: count)) }
+    }
+}
+
+extension String : HTTPResponseBody {
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
+        return try ContiguousArray(utf8).withUnsafeBytes(body)
     }
 }
--- a/Sources/HTTP/HTTPResponse.swift
+++ b/Sources/HTTP/HTTPResponse.swift
@@ -9,6 +9,13 @@
 import Foundation
 import Dispatch
 
+/// HTTP Response for possible use in the client.
+public struct HTTPResponse {
+    public var httpVersion : HTTPVersion
+    public var status: HTTPResponseStatus
+    public var headers: HTTPHeaders
+}
+
 /// Object that code writes the response and response body to. 
 public protocol HTTPResponseWriter : class {
     func writeHeader(status: HTTPResponseStatus, headers: HTTPHeaders, completion: @escaping (Result) -> Void)
--- a/Sources/HTTP/HTTPStreamingParser.swift
+++ b/Sources/HTTP/HTTPStreamingParser.swift
@@ -318,7 +318,7 @@ public class StreamingParser: HTTPResponseWriter {
             return
         }
         
-        var header = "\(parsedHTTPVersion!) \(status.code) \(status.reasonPhrase)\r\n"
+        var header = "HTTP/1.1 \(status.code) \(status.reasonPhrase)\r\n"
 
         let isContinue = status == .continue
 
--- a/Sources/HTTP/HTTPStreamingParser.swift
+++ b/Sources/HTTP/HTTPStreamingParser.swift
@@ -312,99 +312,109 @@ public class StreamingParser: HTTPResponseWriter {
         return HTTPRequest(method: parsedHTTPMethod!, target: parsedURL!, httpVersion: parsedHTTPVersion!, headers: parsedHeaders)
     }
     
-    public func writeContinue(headers: HTTPHeaders?) /* to send an HTTP `100 Continue` */ {
-        var status = "HTTP/1.1 \(HTTPResponseStatus.continue.code) \(HTTPResponseStatus.continue.reasonPhrase)\r\n"
-        if let headers = headers {
-            for (key, value) in headers.makeIterator() {
-                status += "\(key): \(value)\r\n"
-            }
-        }
-        status += "\r\n"
-        
-        // TODO use requested encoding if specified
-        if let data = status.data(using: .utf8) {
-            self.parserConnector?.queueSocketWrite(data)
-        } else {
-            //TODO handle encoding error
-        }
-    }
-    
-    public func writeResponse(_ response: HTTPResponse) {
+    public func writeHeader(status: HTTPResponseStatus, headers: HTTPHeaders, completion: @escaping (Result) -> Void) {
+        // TODO call completion()
         guard !headersWritten else {
             return
         }
         
-        var headers = "HTTP/1.1 \(response.status.code) \(response.status.reasonPhrase)\r\n"
-        
-        switch(response.transferEncoding) {
-        case .chunked:
-            headers += "Transfer-Encoding: chunked\r\n"
-            isChunked = true
-        case .identity(let contentLength):
-            headers += "Content-Length: \(contentLength)\r\n"
+        var header = "\(parsedHTTPVersion!) \(status.code) \(status.reasonPhrase)\r\n"
+
+        let isContinue = status == .continue
+
+        var headers = headers
+        if !isContinue {
+            adjustHeaders(status: status, headers: &headers)
         }
         
-        for (key, value) in response.headers.makeIterator() {
-            headers += "\(key): \(value)\r\n"
+        for (key, value) in headers {
+            // TODO encode value using [RFC5987]
+            header += "\(key): \(value)\r\n"
         }
+        header.append("\r\n")
         
-        let availableConnections = maxRequests - (self.connectionCounter?.connectionCount ?? 0)
-        
-        if  clientRequestedKeepAlive && (availableConnections > 0) {
-            headers.append("Connection: Keep-Alive\r\n")
-            headers.append("Keep-Alive: timeout=\(Int(StreamingParser.keepAliveTimeout)), max=\(availableConnections)\r\n")
-        }
-        else {
-            headers.append("Connection: Close\r\n")
-        }
-        headers.append("\r\n")
-        
+        // FIXME headers are US-ASCII, anything else should be encoded using [RFC5987] some lines above
         // TODO use requested encoding if specified
-        if let data = headers.data(using: .utf8) {
+        if let data = header.data(using: .utf8) {
             self.parserConnector?.queueSocketWrite(data)
-            headersWritten = true
+            if !isContinue {
+                headersWritten = true
+            }
         } else {
             //TODO handle encoding error
         }
     }
+
+    func adjustHeaders(status: HTTPResponseStatus, headers: inout HTTPHeaders) {
+        for header in status.suppressedHeaders {
+            headers[header] = nil
+        }
+
+        if headers[.contentLength] != nil {
+            headers[.transferEncoding] = "identity"
+        } else if parsedHTTPVersion! >= HTTPVersion(major: 1, minor: 1) {
+            switch headers[.transferEncoding] {
+                case .some("identity"): // identity without content-length
+                    clientRequestedKeepAlive = false
+                case .some("chunked"):
+                    isChunked = true
+                default:
+                    isChunked = true
+                    headers[.transferEncoding] = "chunked"
+            }
+        } else {
+            // HTTP 1.0 does not support chunked
+            clientRequestedKeepAlive = false
+            headers[.transferEncoding] = nil
+        }
+
+        let availableConnections = maxRequests - (self.connectionCounter?.connectionCount ?? 0)
+
+        if clientRequestedKeepAlive && (availableConnections > 0) {
+            headers[.connection] = "Keep-Alive"
+            headers[.keepAlive] = "timeout=\(Int(StreamingParser.keepAliveTimeout)), max=\(availableConnections)"
+        } else {
+            headers[.connection] = "Close"
+        }
+    }
     
-    public func writeTrailer(key: String, value: String) {
+    public func writeTrailer(_ trailers: HTTPHeaders, completion: @escaping (Result) -> Void) {
         fatalError("Not implemented")
     }
     
-    public func writeBody(data: DispatchData, completion: @escaping (Result<POSIXError, ()>) -> Void) {
-        writeBody(data: Data(data), completion: completion)
-    }
-    
-    public func writeBody(data: Data, completion: @escaping (Result<POSIXError, ()>) -> Void) {
+    public func writeBody(_ data: UnsafeHTTPResponseBody, completion: @escaping (Result) -> Void) {
         guard headersWritten else {
             //TODO error or default headers?
             return
         }
         
-        guard data.count > 0 else {
-            // TODO fix Result
-            completion(Result(completion: ()))
+        guard data.withUnsafeBytes({ $0.count > 0 }) else {
+            completion(.ok)
             return
         }
         
-        var dataToWrite: Data!
+        let dataToWrite: Data
         if isChunked {
-            let chunkStart = (String(data.count, radix: 16) + "\r\n").data(using: .utf8)!
-            dataToWrite = Data(chunkStart)
-            dataToWrite.append(data)
-            let chunkEnd = "\r\n".data(using: .utf8)!
-            dataToWrite.append(chunkEnd)
+            dataToWrite = data.withUnsafeBytes {
+                let chunkStart = (String($0.count, radix: 16) + "\r\n").data(using: .utf8)!
+                var dataToWrite = chunkStart
+                dataToWrite.append(UnsafeBufferPointer(start: $0.baseAddress?.assumingMemoryBound(to: UInt8.self), count: $0.count))
+                let chunkEnd = "\r\n".data(using: .utf8)!
+                dataToWrite.append(chunkEnd)
+                return dataToWrite
+            }
+        } else if let d = data as? Data {
+            dataToWrite = d
         } else {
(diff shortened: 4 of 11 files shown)


helje5:
> httpVersion ... - a server should use the HTTP version which was requested by a client

This is plain wrong and clarified in [RFC2145](https://www.ietf.org/rfc/rfc2145.txt):

> An HTTP server SHOULD send a response version equal to the highest
   version for which the server is at least conditionally compliant, and
   whose major version is less than or equal to the one received in the
   request

etc. Please read RFCs, they are useful when implementing standards ...

helje5:
> HTTPResponse is redundent

Typo ;-), and well, the original proposal by Johannes didn't actually have it.

I'm not sure what is redundant about it, it is just another way to store the information associated w/ a response? Please elaborate.

Both approaches seem reasonable to me.

aleksey-mashanov:
The meaning of my message does not conflict with RFC2145. I just meant that it is not an application code who must point to used http server which http version to use. It is an agreement between http client and http server which should not be visible to the outer scope.

helje5:
> UnsafeHTTPResponseBody

This allows additional buffer types, but I don't see the gain (at this level) over dispatch-data. It is an extra indirection (the reverse of the proposed 'Simplification') and much worse as it is doesn't support disjunct buffers aka buffer vectors to allow efficient use of `writev` and companions.

Unlike w/ the headers, I'm all against this one. DispatchData is the better option to avoid unnecessary copying and writes. We can talk about more efficient buffer based access (the proposal is certainly not it), but I don't think it is worth it for this effort. DispatchData seems to be a good choice here, reasonable efficiency at a reasonable abstraction level.

P.S.: In general I dislike how multiple things are mixed into a single PR.

helje5:
(Hm, comments do not support threaded responses? That was the whole reason I split up comments :-/ ... so please, a PR per feature ...)

Wrt the version: If the application is responsible for sending the headers, it is also responsible for sending the HTTP version as this defines the meaning of the headers. That is obvious if you think about HTTP/1.0 and HTTP/1.1.

aleksey-mashanov:
> Typo ;-)

Good catch))

> I'm not sure what is redundant about it, it is just another way to store the information associated w/ a response? Please elaborate.

`HTTPResponse` contains only a response status and headers. `Transfer-Encoding` is really one of http headers, I see no reason to honour it so.
`writeResponse` and `writeContinue` have the same meaning, why should they be separated? Is `continue` a response in this case? I think not. Do we really want always to construct http response with status and headers? I think not, in most cases there will be only status.

aleksey-mashanov:
> Wrt the version: If the application is responsible for sending the headers, it is also responsible for sending the HTTP version as this defines the meaning of the headers. That is obvious if you think about HTTP/1.0 and HTTP/1.1.

And does not make sense if you think about HTTP/1.1 and HTTP/2.0. If you don't believe me, you can look here: https://golang.org/pkg/net/http/#ResponseWriter

aleksey-mashanov:
> This allows additional buffer types, but I don't see the gain (at this level) over dispatch-data. It is an extra indirection (the reverse of the proposed 'Simplification') and much worse as it is doesn't support disjunct buffers aka buffer vectors to allow efficient use of writev and companions.

`writev` can be handled by `as?` to the specific protocol or type (similar to `HTTPResponseBody`). May be we should add `withUnsafeBytesVector`?

> Unlike w/ the headers, I'm all against this one. DispatchData is the better option to avoid unnecessary copying and writes. We can talk about more efficient buffer based access (the proposal is certainly not it), but I don't think it is worth it for this effort. DispatchData seems to be a good choice here, reasonable efficiency at a reasonable abstraction level.

`DispatchData` is about Apple world. Server side development is not. In my opinion GCD is not for high load servers handling tenth of thousands queries per second on a commodity hardware. May be I am wrong, I have not use it, but threads are not cheap and I see no reason for them on server where amount of concurrent queries are much higher then amount of cpu cores. Any way I want to use proven solutions for such tasks: coroutines (without threads, using `libcoro`) and asynchronous queries (using `libev` or `libuv` w/o threads also). Why should I use `DispatchData` in this case?

helje5:
> HTTPResponse contains only a response status and headers

Plus the version. Only that triple defines the meaning in full. You can keep it in a structure or in the method signature, but they belong together. Headers and status codes w/o version have ambiguous meaning.

I agree w/ you wrt to Transfer-encoding. In a way this relates to the discussion whether header values should be kept as strings/bytes or as typed entities (HTTPTransferEncoding in this case).

> writeResponse and writeContinue have the same meaning, why should they be separated?

Because they do not have the same meaning. WriteContinue continues an inflight request, while WriteResponse signals response completion (pending/ data).

The implementation may be kinda similar (though it is really not because the continue never carries headers and is optional, right?), the semantics are actually quite different. E.g. there can be multiple 100 calls to keep a connection alive (and may modify socket timeout configuration and such), while there will only ever be a single real response.


>  If you don't believe me

This has absolutely nothing to do w/ 'believing'. I pointed out an RFC which explicitly contradicts your claims and explains in detail why. If there is an updated RFC, please point me to it.

(I can see where you are going, but this is higher level framework thinking. At that HL framework level the decision should be made what an app can and can't do - i.e. a higher level framework may choose to abstract away HTTP protocol versions and not give apps access to arbitrary HTTP headers and such.)


> writev can be handled by as? to the specific protocol or type (similar to HTTPResponseBody). 

Having to resort to as? is awful design and isn't free either. You introduce a new layer to just as? around it? ;-)


> May be we should add withUnsafeBytesVector?

Well, I don't think the protocol is desirable in the first place as it has overhead and we already have an accepted and viable concrete implementation (DispatchData).

If other people agree that this protocol should really exist, yes, maybe an withUnsafeBytesVector should replace the withUnsafeBytes version (definitely not add it, it is a low level primitive and the vector variant fully covers the non-vector case).

> DispatchData is about Apple world

Not more than Foundation. Dispatch is part of Swift. I'd prefer if DispatchData would be deprecated in favour of Data, but AFAIK Data (unfortunately) moved into a different direction, being a plain in-memory buffer. (Maybe that can be fixed.?)

Your talking about GCD has nothing to do with DispatchData (and is again plain wrong, GCD doesn't require threads for I/O and is conceptually the same like libuv, but that doesn't matter for the discussion, lets stay on topic).

> Why should I use DispatchData in this case.

Because it is standard part in Swift and provides a solution for passing around disjoint buffers of data.

aleksey-mashanov:
> Headers and status codes w/o version have ambiguous meaning.

Agreed. But http server better knows standards then any higher level application.

> continue never carries headers

Please, point me to an RFC. In an initial implementation a possibility to pass headers with continue was exist so it was kept.

> (I can see where you are going, but this is higher level framework thinking. At that HL framework level the decision should be made what an app can and can't do - i.e. a higher level framework may choose to abstract away HTTP protocol versions and not give apps access to arbitrary HTTP headers and such.)

May be my understanding of this project differs from yours. The only thing I want is to have a universal API between application and any http server (uwsgi, nginx, apache, etc...). If connection is already upgraded to 2.0 by an underlying server why should I (or any higher level framework) handle this?

> Not more than Foundation.

I hate it. Especially on Linux. Sorry.

> Because it is standard part in Swift and provides a solution for passing around disjoint buffers of data.

Mostly it will be strings converted to `DispatchData` with copying. Or encoded JSON. Is it `Data` or `DispatchData` in Swift 4.0? Swift has no standard data buffers. Anything except standard library is de-facto non-standard and not proven to work well on Linux. The best solution it has is `UnsafeRawBufferPointer`.

aleksey-mashanov:
> GCD doesn't require threads for I/O and is conceptually the same like libuv

Hmmm... Apple documentation misleaded me. `libkqueue` is ok.

helje5:
>> Headers and status codes w/o version have ambiguous meaning.

> Agreed. But http server better knows standards then any higher level application.

So we agree that it doesn't make sense that expose headers w/o a version. Good.
We seem to disagree on whether this API is a higher level application API. But I think there was definitive agreement in the group that this API is specifically *not* a higher level application API.

>> continue never carries headers

> Please, point me to an RFC. In an initial implementation a possibility to pass headers with continue was exist so it was kept.

Seriously? [RFC2616](https://www.ietf.org/rfc/rfc2616.txt).

> May be my understanding of this project differs from yours. The only thing I want is to have a universal API between application and any http server (uwsgi, nginx, apache, etc...). If connection is already upgraded to 2.0 by an underlying server why should I (or any higher level framework) handle this?

I'm not sure why you think that. What does the upgrading have to do with it? We agreed above that the meaning of header values is established by the HTTP version. So anytime you use header values, you need to provide your assumption of the HTTP version they belong to. If the server part of the API wants to rewrite that for a different version, it is free to do that?

Now we can also fix the semantic of the headers of the API to a specific version (say HTTP/1.1 [and I think 2.0 didn't change it, not 100% sure]). But I'm not sure whether that is a good idea (at least not at this level of the API).

> I hate it.

I'm not so sure whether it is sensible to base discussions on 'hate' :->

helje5:
> Hmmm... Apple documentation misleaded me. libkqueue is ok.

Yes, GCD uses kqueue for I/O, but that has zarro relevance for the discussion, could you please stay on topic?

The thing discussed was DispatchData which is just an object supporting the needs for efficient async I/O, it has nothing to do with I/O scheduling. At all.
(Also, it is fully supported on Linux and well tested.)

aleksey-mashanov:
> Seriously? RFC2616.

Still don't see. Quote please. The RFC is very long.

> I'm not sure why you think that. What does the upgrading have to do with it?

I meant that underlying implementation of a server usually will be smart enough.

> But I think there was definitive agreement in the group that this API is specifically not a higher level application API

uWSGI, PSGI, CGI, FastCGI - are they too high?

seabaylea:
I agree with @helje5 that breaking this down into small issues/PRs will make it much easier to move things forward. Below are comments on some of the areas covered:

* Removal of HTTPResponse
We certainly need a HTTPResponse type, because at some point we'll definitely want to implement HTTP client requests as well. I agree that for server responses we're just wrapping in a type where function arguments are sufficient.

* transferEncoding
Having transferEncoding as its own field in HTTPResponse makes little sense as it is really a header. Having said that, keeping it as a separate function argument probably makes sense because the response does need to be identity or chunked (the other flags are additive).

* HTTP Version
I agree that there shouldn't be a need to explicitly set the version for the HTTP response - the parser already knows what it needs to be -  equally the responding application code can knows what it going to be (by inspecting the request), so I'm not sure where the ambiguity would be.

helje5:
> HTTP Version

Chris, this is covered in [RFC 2145](https://www.ietf.org/rfc/rfc2145.txt). Don't be afraid, it is a super-duper short RFC ;-)
There is a reason why both, the request and the response have to state the version they use ...

_Right now_ it may not look very important anymore because the transition from HTTP 0.9 and 1.0 to 1.1 is essentially done and HTTP/2.0 AFAIK didn't touch the meaning of headers in relevant ways.
However, there could be a HTTP/1.2 or HTTP/2.1 which makes this necessary again.

It's part of the protocol and should be exposed both ways.

aleksey-mashanov:
> I agree with @helje5 that breaking this down into small issues/PRs will make it much easier to move things forward.

Agree too. The problem that stopped me from the split is that concurrent PRs will lead to a complex merge with a lot of conflicts. Making all of this changes sequentially will be a long story.

> We certainly need a HTTPResponse type, because at some point we'll definitely want to implement HTTP client requests as well.

I'd prefer symmetrical API for request and response, client and server too. But in this case there will be a lot of *notes* about that to do with this object on client and that on server. You can see this badness in [golang request](https://golang.org/pkg/net/http/#Request). And this example is just synchronous API, with async things are more complicated. node.js is on the other side: it groups client response with server request calling it [IncomingMessage](https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_class_http_incomingmessage).

For this reasons, I think that it is better to provide separate client/server request/response. This will be four interfaces but they will allow only operations/options they are capable to provide to.

> Having said that, keeping it as a separate function argument probably makes sense because the response does need to be identity or chunked (the other flags are additive).

Almost always application code does not bother about which one (chunked or identity) will be used. It just provide `Content-Length` header (then identity is used) or not (then chunked is used or connection is closed, see `adjustHeaders` method in this PR). And it is not simple to application code to choose which one to use. It depends on http protocol version, capabilities set by client in `TE` header and `Content-Length` header. I don't think anybody outside http server implementation wants to think about this every time they construct a response.

seabaylea:
> I'd prefer symmetrical API for request and response, client and server too. But in this case there will be a lot of notes about that to do with this object on client and that on server. You can see this badness in golang request.

Ideally the types should be symmetric, but the access to them wouldn't be, i.e in the case of the server you have a HTTPRequest and a ResponseWriter, in the case of the client the equivalent would be to have a RequestWriter and a HTTPResponse.
Are the issues with the golang API related it not actually being symmetric? (i.e., the client API has direct access to the Request object).

> Almost always application code does not bother about which one (chunked or identity) will be used. It just provide Content-Length header (then identity is used) or not (then chunked is used or connection is closed, see adjustHeaders method in this PR).

This could be dealt with by having default parameters in the API, for example (under the existing code):
```swift
public struct HTTPResponse {    
    public init (httpVersion: HTTPVersion, status: HTTPResponseStatus, transferEncoding: HTTPTransferEncoding = .chunked , headers: HTTPHeaders) { }
}
```
which drops the need to provide a transferEncoding unless you want to use identity with a provided Content-Length.

seabaylea:
@helje5 I assume the relevant sections are:
> An HTTP server SHOULD send a response version equal to the highest version for which the server is at least conditionally compliant, and whose major version is less than or equal to the one received in the request.

and 

> An HTTP server MAY send a lower response version, if it is known or suspected that the client incorrectly implements the HTTP specification, but this should not be the default, and this SHOULD NOT be done if the request version is HTTP/1.1 or greater.

i.e. its acceptable send a response with a version lower then the request.

Is it reasonable to default the version of the response to the version of the request (and provide the ability to override)?

aleksey-mashanov:
Some more words about http version of a response.

The key question we should answer is *who is responsible for choosing correct version of a response: a low-level http server bellow the API or high-level framework above*. Any of this levels can follow RFC2145 to correctly set version, but who of them has more abilities and rights to do this?

The second question is: *who is responsible to handle http protocol upgrade to HTTP/2.0*. Here it shouldn't be forgiven that there is a [difference](https://tools.ietf.org/html/rfc7540#section-3.3) for http and https.

The third question: *what to do with a plenty of proven http servers which don't provide an ability to set a response version?* Should they ignore provided version or should we throw them away?

And, last but not least, Swift is not the first and not the last language with such intermediate API between http server and web framework. I looked at some popular modern APIs (Python/WSGI, Ruby/Rack, Perl/PSGI) and on some http server implementations (golang, node.js). None of them have ability to control version of a http response. May be for some reasons.

helje5:
@seabaylea I read the first quote so that the server can reply with 1.4 to a 1.1 request. (only the major needs to match, and the minor should send the highest it supports at least in some way).

> Is it reasonable to default the version of the response to the version of the request.

I'm not sure, I would say the answer is "no". Why? Because the application code setting headers needs to know whether the headers work for the specific version (and how). If they don't specify the version, they can't know what to expect (as it is arbitrary and decided by the client).

For a reasonable default I think a better way would be to fix the version of the response to 1.1/2.0 (and therefore fix the semantics of the headers to that version). What I said before/above:

> Now we can also fix the semantic of the headers of the API to a specific version (say HTTP/1.1 [and I think 2.0 didn't change it, not 100% sure]). But I'm not sure whether that is a good idea (at least not at this level of the API).

(a version-aware app could then still change the version and do version specific stuff, but the default would be 1.1 semantics)

My opinion is still that this doesn't belong into this low-level API. A higher level API can abstract stuff like that away from the application developer and make stuff nice and restrict the user to sensible defaults (i.e. not allow him to set arbitrary headers).

BTW: What about about upgrade? (WebSockets, SSL proxying, etc).

helje5:
@aleksey-mashanov What we are discussing here is the Swift HTTP API, not some gateway protocol that is a subset of HTTP. This is all about real, frontend HTTP servers.

aleksey-mashanov:
> Are the issues with the golang API related it not actually being symmetric?

The answer is in the comments to specification to Request type here: https://golang.org/pkg/net/http/#Request
Almost every method has two comments: one for server and one for client.

In my practice client and server (irrespective to Go and HTTP) have different use cases and seldom can be covered by the same implementation. For server request is incoming and response is outgoing, for client vice versa. For server response should be easy to construct, for client - to read from. There are a lot of small differences which should be handled differently. Anyway, some their parts are definitely must be the same: method, status, headers, may be uri.

helje5:
> Almost every method has two comments: one for server and one for client.

Yes, we should absolutely avoid that non-sense. The request URI field is absolutely the same whether it is coming from the client or the server (the client can send it and the server can receive it, hu?). This shows how bad it is:

> For client requests, the URL's Host specifies the server to connect to

They overload the field to have different meanings depending on the context an object is used?! How do I create a proxy request using that API if the URL in the request designates the connect host? How do I make an OPTIONS request to discover server caps? Common ... This is *really* bad stuff.

*Please* lets not replicate API which has two meanings for every method ... :-)

aleksey-mashanov:
> What we are discussing here is the Swift HTTP API, not some gateway protocol that is a subset of HTTP. This is all about real, frontend HTTP servers.

The purpose of any API is unification, any unification is a subset of possibilities. Too much possibilities to higher level leads to too much responsibility. Usability is all about limiting responsibilities for simplicity.

There are some best practices from the long history of web development. Gateway APIs is one of them adopted by many servers. Another one is application servers (apache, uwsgi) which are used by folks for many reasons. And don't forget about *nginx* - usually it is used as a frontend routing server with a backend application servers behind it. A protocol between a frontend server and a backend server on the network is not always just plain HTTP (it can be *uwsgi* for performance reasons) but anyway is HTTP logically.

Providing too much possibilities to higher level limits amount of http server implementations which can be used bellow without a lot of notes and warnings about that shouldn't be done for compatibility reasons.

aleksey-mashanov:
> If they don't specify the version, they can't know what to expect (as it is arbitrary and decided by the client).

It is the destiny of a server: to do that a client asks. Not vice versa.

aleksey-mashanov:
> How do I create a proxy request using that API if the URL in the request designates the connect host? How do I make an OPTIONS request to discover server caps?

It looks ugly, but it is possible and this approach is widely used and de-facto standard. For instance, `curl` from command-line works the same way: you can setup IP to connect to in URI and server hostname in `Host` header. And there are no problems with proxies.

aleksey-mashanov:
> `transferEncoding: HTTPTransferEncoding = .chunked`

`chunked` is not the best default. It is better to use `identity` whenever it is possible because it allows client to allocate exactly one buffer of known size and to reject queries with size out of a limit. The same can be achieved using`contentLength: Int? = nil` but for this relatively low level API it is much simpler just to set header.

seabaylea:
>> Is it reasonable to default the version of the response to the version of the request.
> I'm not sure, I would say the answer is "no". Why? Because the application code setting headers needs to know whether the headers work for the specific version (and how). If they don't specify the version, they can't know what to expect (as it is arbitrary and decided by the client).

> For a reasonable default I think a better way would be to fix the version of the response to 1.1/2.0 (and therefore fix the semantics of the headers to that version). 

Right - so in our current case (without 2.0 support), the default response would be 1.1. With HTTP/2 support added the response would default to 1.1 or 2.0 dependent on the request. This is really just providing convenience so an application or framework writer doesn't have explicitly specify the response version, whilst retaining the ability to do so.

>> Are the issues with the golang API related it not actually being symmetric?
> The answer is in the comments to specification to Request type here: https://golang.org/pkg/net/http/#Request
Almost every method has two comments: one for server and one for client. 

Isn't this because Golang makes Request directly accessible to both the client and server APIs:
Client:
```
    func (c *Client) Do(req *Request) (*Response, error)
    func (c *Client) Get(url string) (resp *Response, err error)
    func (c *Client) Head(url string) (resp *Response, err error)
```
Server:
```
func HandleFunc(pattern string, handler func(ResponseWriter, *Request))
```
It doesn't look like the same issues exist for Response, which is only directly accessible to the client.

>> If we implement the client to have an API to create the request and/or a request writer, rather than direct access to the Request type itself, do these issues go away?
```
transferEncoding: HTTPTransferEncoding = .chunked
```
> chunked is not the best default. It is better to use identity whenever it is possible because it allows client to allocate exactly one buffer of known size and to reject queries with size out of a limit. The same can be achieved usingcontentLength: Int? = nil but for this relatively low level API it is much simpler just to set header.

Right - so what I meant here is that the application or framework could make either of the following calls:
Use the default (`.chunked`):
```swift
res.writeResponse(HTTPResponse(httpVersion: req.httpVersion,
                                               status: .ok,
                                               headers: [:]))
```
Use identity with a size 16:
```swift
 res.writeResponse(HTTPResponse(httpVersion: req.httpVersion,
                                               status: .ok,
                                               transferEncoding: .identity(contentLength: 16)
                                               headers: [:]))
```

aleksey-mashanov:
> With HTTP/2 support added the response would default to 1.1 or 2.0 dependent on the request.

Response constructor has no request parameter right now. This can be easily solved in another way by adding `var httpVersion: HTTPVersion { get }` to `HTTPResponseWriter`. I still recommend to not add `set` capability, only `get`. HTTP version has more ties with connection then headers. I hardly can imagine http server implementation which will correctly support changing of response version without limitation.

Answering with a last minor supported by a server for the same major given by a client has sense. But this should be done with care about actual client version given in the request. For example, where is the note in RFC7230 about chunked transfer encoding:

```
   Transfer-Encoding was added in HTTP/1.1.  It is generally assumed
   that implementations advertising only HTTP/1.0 support will not
   understand how to process a transfer-encoded payload.  A client MUST
   NOT send a request containing Transfer-Encoding unless it knows the
   server will handle HTTP/1.1 (or later) requests; such knowledge might
   be in the form of specific user configuration or by remembering the
   version of a prior received response.  A server MUST NOT send a
   response containing Transfer-Encoding unless the corresponding
   request indicates HTTP/1.1 (or later).
```

> `transferEncoding: .identity(contentLength: 16)`

Identity transfer encoding does not require `Content-Length` header. Instead `Connection: close` can be used to indicate end of response.
`Transfer-Encoding` header has more then two possible values. `compress, gzip, deflate` are widely used in modern web to reduce mobile traffic.

seabaylea:
>>With HTTP/2 support added the response would default to 1.1 or 2.0 dependent on the request.

> Response constructor has no request parameter right now. 

It doesn't, but the responder protocol/closure definition does provide access to the request:
```swift
func serve(req: HTTPRequest, res: HTTPResponseWriter ) -> HTTPBodyProcessing { }
```
so the application/framework developer has access to the request when building the response if they want to set a version explicitly, e.g.:
```swift
    func serve(req: HTTPRequest, res: HTTPResponseWriter ) -> HTTPBodyProcessing {
	        res.writeResponse(HTTPResponse(httpVersion: req.httpVersion,
	                                       status: .ok,
	                                       transferEncoding: .chunked,
	                                       headers: [:]))
	        return .discardBody
	    }
```
and if they don't the HTTPStreamingParser can set the HTTPResponse version based on the incoming request.

>>transferEncoding: .identity(contentLength: 16)

>Identity transfer encoding does not require Content-Length header. Instead Connection: close can be used to indicate end of response.
Transfer-Encoding header has more then two possible values. compress, gzip, deflate are widely used in modern web to reduce mobile traffic.

We could modify the transferEncoding enum to be:
```swift
public enum HTTPTransferEncoding {
    case identity                // use identify plus close to generate length
    case contentLength    // use identity plus specified length
    case chunked              // use chunked
}
```
to give a way of specifying the use of close or Content-Length. Alternatively we could just remove the parameter from `identity` and use the presence or absence of the Content-Length header (and leave it to any framework to make the API more usable for the developer).

The other available values: `compress, gzip, deflate` are additive and I would expect would be directly set through the headers by higher level frameworks etc (e.g. when using compression middlewares)

helje5:
Why can't you just leave the transfer-encoding as a regular header? The response writer can inspect them, check whether a TE or CL is set and do the right thing.

```swift
func serve(req: HTTPRequest, res: HTTPResponseWriter ) -> HTTPBodyProcessing {
  res.writeResponse(HTTPResponse(httpVersion: req.httpVersion,
	                           status: .ok,
                                   headers: ["Transfer-Encoding": "chunked"]))
  return .discardBody
}
```

aleksey-mashanov:
Once again about http version, headers and status.

In general, HTTP headers are independent from HTTP version (see [RFC3864](https://tools.ietf.org/html/rfc3864)) and registered by IANA: https://www.iana.org/assignments/message-headers/message-headers.xhtml. Quite a few of them are defined in RFCs related to HTTP/1.1 but this doesn't mean that these headers are tied to HTTP/1.1. One more example is [CORS](https://www.w3.org/TR/cors/) specification. It defines several widely used headers without any word about HTTP version.

HTTP status codes and their meaning are also registered by IANA: https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

helje5:
Yup, as covered in RFC 2145. The HTTP version defines the layout/transport/connection-handling of the HTTP messages, i.e. headers related to the _HTTP protocol itself_, this is not about user level headers. TE/chunked is a good example of an affected header (as mentioned in the RFC), and is also a good example for a potential HTTP version change in the next years (there could be new TEs which renders the protocol incompatible if not done right).

I think if we default to 1.1/2.0 by default we should be good.

seabaylea:
> * HTTPResponse is redundent. Its status and headers properties
are moved to parameters of writeHeader method of HTTPResponseWriter
instance. 

I agree that HTTPResponse should be redundant for the server case, but not for the client so we should keep the definition but remove it from the server API.

> * transferEncoding handling is moved to HTTPHeaders - it
is not as simple as an enum and depends on used HTTP protocol version
and capabilites set by a client in its request headers.

Agreed. 

> * httpVersion should not be configured by an outer scope - a server should
use the HTTP version which was requested by a client.

As discussed, this should now default to 1.1 or 2.0 as appropriate.

> * writeResponse and writeContinue are replaced by writeHeader.

FYI, I can't find anything in the RFC that states whether Continue can/cannot include headers, equally I'm not sure that the client would expect to receive headers in the response. If we're not going to support writing headers with continue, then retaining the writeContinue call makes sense.

> * writeTrailer accepts trailers as HTTPHeaders struct.

Agreed.

> * All write* methods accept a completion handler.

Agreed.

> * writeBody accepts UnsafeHTTPResponseBody instead of just Data
and Dispatch. It should be used only in the scope of a call and a
copy should be made or value should be checked for HTTPResponseBody
conformance before storing it anythere. This approch allows to reduce
memory copying.

Don't we get this from DispatchData?

> * Result<POSIXError, Void> enum is replaced with Result. In the
first case POSIXError and Void are generic parameters names, not
types. I think this behaviour was unintended.

Agreed

aleksey-mashanov:
> I agree that HTTPResponse should be redundant for the server case, but not for the client so we should keep the definition but remove it from the server API.

May be it should be moved somewhere to not confuse with server?

> FYI, I can't find anything in the RFC that states whether Continue can/cannot include headers, equally I'm not sure that the client would expect to receive headers in the response.

I didn't find anything too.

> If we're not going to support writing headers with continue, then retaining the writeContinue call makes sense.

In the previous version `writeContinue` supported writing headers and had the following definition: `func writeContinue(headers: HTTPHeaders?)`. May be we should just remove dubious API and allow higher level code to handle this situation with `writeHeader`?

> Don't we get this from DispatchData?

From `DispatchData` we get no-copy for `DispatchData` only. In the majority of cases this will be `String` or `Data` instead.

helje5:
> FYI, I can't find anything in the RFC that states whether Continue can/cannot include headers

I couldn't find a definitive answer to that in the RFC either. What would they contain, the request wasn't even fully sent yet? The only thing I can think of is maybe Server or Date.

The purpose of Expect-100 continue is to check authentication and general applicability before a potentially large request body is sent over (say an image or full calendar upload). The answer to this is either 'yes' or, the final response.

helje5:
> From DispatchData we get no-copy for DispatchData only.

Hu? When was that removed? The API I know supports arbitrary sources, i.e. `init(bytesNoCopy:deallocator:)`

seabaylea:
>> I agree that HTTPResponse should be redundant for the server case, but not for the client so we should keep the definition but remove it from the server API.

> May be it should be moved somewhere to not confuse with server?

Its already in its own source file, so lets leave it there. Lets also remove it from the API.md file though as it won't be part of the API (until we extend it for client support).

>> If we're not going to support writing headers with continue, then retaining the writeContinue call makes sense.

> In the previous version writeContinue supported writing headers and had the following definition: func writeContinue(headers: HTTPHeaders?). May be we should just remove dubious API and allow higher level code to handle this situation with writeHeader?

As (I think) we're saying it doesn't make sense to be able to pass headers, we should have a writeContinue() convenience function, especially as otherwise you'd have to use writeHeader, which expects headers to be passed.

>> Don't we get this from DispatchData?

> From DispatchData we get no-copy for DispatchData only. In the majority of cases this will be String or Data instead.

Are we now agreed to leave the use of DispatchData as-is?

aleksey-mashanov:
> The purpose of Expect-100 continue is to check authentication...

In this case server could return authentication token. It is unlikely case anyway because I see no way to handle 100 answer from a server in a browser using `XMLHTTPRequest` (browser will handle it behind the scene).

> Hu? When was that removed? The API I know supports arbitrary sources, i.e. `init(bytesNoCopy:deallocator:)`

OK. And how you will pass `String` using this constructor?

> ...as otherwise you'd have to use writeHeader, which expects headers to be passed.

No, you can use `wirteHeader(status:)` defined in an extension:

```
public func writeHeader(status: HTTPResponseStatus)
```

helje5:
>> The purpose of Expect-100 continue is to check authentication...

> In this case server could return authentication token.

I honestly have no idea what are you talking about? The client sends a Expect-100 continue to check whether the server is going to accept the request before sending the whole payload. The server sends a 100 continue and only then the client will start transmitting the request body. If the server knows it is going to reject the request, it is immediately going to send a 401, 403 etc, whatever makes sense.

XMLHttpRequest doesn't expose 100-continue (it may use it internally, but I guess this may be rather unlikely). Yet a significant amount of server side Swift traffic will be driven by app requests, not just the web ...

Again, the significant difference between writeContinue() and writeHeader() is not the wire format, but that writeHeader concludes a HTTP transaction and can only be called once, while writeContinue doesn't and keeps the TX open. With all the implications for keep-alive, logging etc.
I.e. calling writeHeader twice should actually throw an error.

aleksey-mashanov:
> I honestly have no idea what are you talking about? ... Yet a significant amount of server side Swift traffic will be driven by app requests, not just the web ...

I meant that server could return any non-standard header. The most probable one is access token. But I really don't think it is a good design as authentication and any other activity should be handled separately.

> Again, the significant difference between writeContinue() and writeHeader() is not the wire format, but that writeHeader concludes a HTTP transaction and can only be called once, while writeContinue doesn't and keeps the TX open. With all the implications for keep-alive, logging etc.
I.e. calling writeHeader twice should actually throw an error.

Yes. But does it really means that these must be separate methods? Why not to handle all of this stuff using condition in `writeHeader` implementation?

On the other side some implementations just send `100 Continue` under the hood at the time when the first `read` on the request body is called. I tend to think that it is the best solution.

seabaylea:
> Yes. But does it really means that these must be separate methods? Why not to handle all of this stuff using condition in writeHeader implementation?

Usually, because its less obvious to the user what the behaviour is - if you find that the documentation for the API becomes complex (and starts to look like a switch statement) then we should most likely break the API down.

aleksey-mashanov:
>> Yes. But does it really means that these must be separate methods? Why not to handle all of this stuff using condition in writeHeader implementation?

> Usually, because its less obvious to the user what the behaviour is - if you find that the documentation for the API becomes complex (and starts to look like a switch statement) then we should most likely break the API down.

In general, yes. But in this certain case we will have two possible ways to send `100 Continue`: using `writeContinue` and using `writeHeader(status: .continue)`.  In absence of exact guidelines the last one will not work as expected but for me it is more intuitive.

Anyway I'd prefer the following solution:
* allow user to call `writeHeader(status: .continue)` with predictable result.
* send `100 Continue` automatically on the first read from the body handle if it wasn't sent yet and `Expect: 100-continue` exists in the request header.

Personally I dislike honouring several headers and statuses and giving them separate methods in the general low-level API.

seabaylea:
If we provide a `writeContinue()` call then users are less likely to call `writeHeader(status:, headers:)` won't provide the behaviour that the user might expect - although I agree that providing the `writeContinue()` call doesn't prevent that anyway.

If we provide an automatic `100 Continue` if a `.processBody()` callback is provided, isn't the point moot though? i.e., the user shouldn't need to manually return `.continue`

helje5:
No the point is not mood. Whether or not a request makes sense (and whether the client should continue w/ an upload) needs to be decided in user code (/ higher level framework code). Authentication is the most common thing, but there are other things the server might want to check (e.g. quotas, supported type constraints and such).
A more obscure use is to keep the HTTP connection alive (if the server is working and resets timeouts by sending 100's).

(P.S.: This is not just theoretical stuff, 100 is actually deployed in large scale cloud environments and clients)

aleksey-mashanov:
> Whether or not a request makes sense (and whether the client should continue w/ an upload) needs to be decided in user code (/ higher level framework code).

Absence of an explicit method to send `100 Continue` doesn't mean that a higher level code has no control. If user framework wants to handle the request (quotas, constraints and other stuff is ok) it just reads the request body (in this case `100 Continue` is automatically sent by a lower level part), if it doesn't want - it just ends processing of the request with an appropriate status without reading the body (no `100 Continue`).

seabaylea:
In order to move this forward - there a number of things in the PR we've come to agreement on - why don't we "resolve" the outstanding items by:
1) Removing the explicit `writeContinue()` call - we can always add the API in the future (its harder to remove APIs than add them)
2) Requiring the application/framework code to explicitly respond with `100 Continue` - we can debate adding implicit support in a separate PR where we can also test/show how it works.

helje5:
Both could be simply summarised as: This effort is not going to support 100-continue (at least not now).
100-continue w/o user code involvement makes zarro sense, we can as well drop it then (yes, the framework can just send a 100-continue, but there is a _reason_ why the client sends the Expect in the first place ... 🙄 Expect: 100 w/o actual support in the server only increases latency at no gain).

IMHO, this whole PR should be dropped altogether in favour of new, focused PRs with specific goals. If there are "a number of things in the PR we've come to agreement on", it should be easy to get them through.

P.S.: The thing invalidating this PR the most (IMO) is that just one of the core stakeholders commented on it, the requester being the only other even involved. There must be a reason for this ... (I suppose the broad nature of the PR is probably one of them)

aleksey-mashanov:
I think this project is not in the state where rejecting is better then moving forward. It is better to do something wrong then doing nothing. This PR is a small step in the right direction. A little of it will be in the final version. For example as I stated above this API will not work with synchronous flow at all. May be it should be split in sync/async API. This API doesn't support passing filehandles (and so `sendfile`). A lot should be done before this project will be usable.

Of course we can play in splitting PRs and discussing once again, but this will be a waste of time.

aleksey-mashanov:
> we can always add the API in the future (its harder to remove APIs than add them)

Definitely! This is the way APIs should be designed.

seabaylea:
@aleksey-mashanov can you update the PR so it reflects what's been discussed? We'll then merge it so we can move forward, and hopefully then work on smaller items at a time.

aleksey-mashanov:
Done

Now PR should be in the state we are agreed on. Except for `UnsafeHTTPResponseBody`. I leaved it untouched because I see no prove of possibility to use `DispatchData` to pass `String` and `Data` without copy in this thread. Anyway it should be extended and polished in following PRs.

seabaylea:
I'm going to merge this to close this one down - the ongoing discussion about the use of DispatchData can be followed up in a separate item.

carlbrown:
Ok, so we've orphaned the `HTTPResponse` struct now? It still exists, but isn't used in anything?  Am I the only one that finds that confusing?

PR #21, opened 2017-07-11
Fix for exclusive access issue in Swift 4

carlbrown:
Swift4's ownership checking doesn't like the way I was accessing `&httpParser` in multiple callbacks.

Fix should be backwards compatible with Swift 3.1.1


seabaylea:
LGTM. Merging.

PR #23, opened 2017-07-18
Perf: read connectionCount only when needed

seabaylea:
`connectionCount` is only needed if keep alive is requested, so accessing the value should be done inside the `clientRequestedKeepAlive` check

### DIFF
--- a/Sources/HTTP/HTTPStreamingParser.swift
+++ b/Sources/HTTP/HTTPStreamingParser.swift
@@ -374,14 +374,16 @@ public class StreamingParser: HTTPResponseWriter {
             headers[.transferEncoding] = nil
         }
 
-        let availableConnections = maxRequests - (self.connectionCounter?.connectionCount ?? 0)
 
-        if clientRequestedKeepAlive && (availableConnections > 0) {
-            headers[.connection] = "Keep-Alive"
-            headers[.keepAlive] = "timeout=\(Int(StreamingParser.keepAliveTimeout)), max=\(availableConnections)"
-        } else {
-            headers[.connection] = "Close"
+        if clientRequestedKeepAlive {
+            let availableConnections = maxRequests - (self.connectionCounter?.connectionCount ?? 0)
+            if availableConnections > 0 {
+                headers[.connection] = "Keep-Alive"
+                headers[.keepAlive] = "timeout=\(Int(StreamingParser.keepAliveTimeout)), max=\(availableConnections)"
+                return
+            }
         }
+        headers[.connection] = "Close"
     }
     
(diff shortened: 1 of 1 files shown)


Sources/HTTP/HTTPStreamingParser.swift
```diff
@@ -374,14 +374,16 @@ public class StreamingParser: HTTPResponseWriter {
...
-        let availableConnections = maxRequests - (self.connectionCounter?.connectionCount ?? 0)
 
-        if clientRequestedKeepAlive && (availableConnections > 0) {
-            headers[.connection] = "Keep-Alive"
-            headers[.keepAlive] = "timeout=\(Int(StreamingParser.keepAliveTimeout)), max=\(availableConnections)"
-        } else {
-            headers[.connection] = "Close"
+        if clientRequestedKeepAlive {
+            let availableConnections = maxRequests - (self.connectionCounter?.connectionCount ?? 0)
+            if (availableConnections > 0) {
```
ianpartridge: Remove parentheses.

Sources/HTTP/HTTPStreamingParser.swift
```diff
@@ -374,14 +374,16 @@ public class StreamingParser: HTTPResponseWriter {
...
+                headers[.connection] = "Keep-Alive"
+                headers[.keepAlive] = "timeout=\(Int(StreamingParser.keepAliveTimeout)), max=\(availableConnections)"
+                return
+            }
         }
+        headers[.connection] = "Close"
```
ianpartridge, line 386: Put this in an `else` and remove the `return` line above.

seabaylea: Unfortunately (I think) what your proposing breaks the original logic flow.

ianpartridge: I don't think it does... Which combination of `clientRequestedKeepAlive` and `availableConnections` do you think would give an incorrect result?

ianpartridge: Oh I see, if we request keepalive but there are no available connections.

Sources/HTTP/HTTPStreamingParser.swift
seabaylea: Updated

seabaylea:
@ianpartridge I've made some updates. Can you take another look?

ianpartridge:
Looks fine now. Does this make a big impact to perf?

seabaylea:
Its noticeable for the non-keep alive case because `connectionCount` is under a lock. I think there's a lot of scope to optimise in this area.

seabaylea:
Merging based on the review from @ianpartridge

PR #27, opened 2017-08-22
Remove HTTPResponseStatus.init(code:reasonPhrase:)

ianpartridge:
This public API seems unwise because it allows someone to create an `HTTPResponseStatus` with an incorrect reason phrase, for example `HTTPResponseStatus(code: 404, reasonPhrase: "OK")`.

### DIFF
--- a/Sources/HTTP/HTTPResponse.swift
+++ b/Sources/HTTP/HTTPResponse.swift
@@ -89,18 +89,11 @@ public struct HTTPResponseStatus: Equatable, CustomStringConvertible, Expressibl
     public let reasonPhrase: String
 
     /// Creates an HTTP response status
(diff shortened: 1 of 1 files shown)


aleksey-mashanov:
Creating response with a custom reason phrase is allowed by [RFC7231](https://tools.ietf.org/html/rfc7231#section-6.1):
> The status codes listed below are defined in this specification,
   Section 4 of [RFC7232], Section 4 of [RFC7233], and Section 3 of
   [RFC7235].  The reason phrases listed here are only recommendations
   -- they can be replaced by local equivalents without affecting the
   protocol.

ianpartridge:
Thanks, I wasn't aware of this.

PR #32, opened 2017-08-23
Remove CHTTPParser dependency, add target

ianpartridge:
Remove the SPM dependency on https://github.com/IBM-Swift/CHTTPParser and bring the C HTTP parser code into this project, as a new target.


seabaylea:
Merging. Note that its also possible to remove the reliance of the `utils.*` files, so that should be done in a followup PR (see #25 from @tanner0101 )

PR #38, opened 2017-08-28
Remove dangling reference left over from rename in PR #26

carlbrown:
This enables the tests to run in Linux again


seabaylea:
LGTM. Merging.

PR #41, opened 2017-09-07
Correct function of Keep-Alive header

djones6:
The HTTP `Keep-Alive` response header \[[1](https://tools.ietf.org/id/draft-thomson-hybi-http-timeout-03.html#rfc.section.2)\] `max` parameter indicates the number of remaining requests that a client may make before the server closes the connection.

The current implementation seems to be trying to take into account the number of active connections to the server when responding, which does not seem correct. This PR changes the value sent to be a simple count of requests remaining.  Once the count becomes zero, a `Connection: Close` header is sent.

\[1\]: https://tools.ietf.org/id/draft-thomson-hybi-http-timeout-03.html#rfc.section.2

### DIFF
--- a/Sources/HTTP/HTTPStreamingParser.swift
+++ b/Sources/HTTP/HTTPStreamingParser.swift
@@ -19,7 +19,7 @@ public class StreamingParser: HTTPResponseWriter {
     /// Time to leave socket open waiting for next request to start
     public static let keepAliveTimeout: TimeInterval = 5
 
-    /// Flag to track if the client wants to send multiple requests on the same TCP connection
+    /// Flag to track if the client wants to send consecutive requests on the same TCP connection
     var clientRequestedKeepAlive = false
 
     /// Tracks when socket should be closed. Needs to have a lock, since it's updated often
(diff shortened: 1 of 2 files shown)


djones6:
@carlbrown Please could you review?  I'd like to understand your original intention w.r.t. connection count.  If the intention is to limit the number of concurrent requests being handled by the server, I would suggest that instead we instead limit the number of connections at the point of acceptance.

carlbrown:
@djones6 Section 2.2.1 of the RFC you included says the `max` parameter is deprecated anyway. Why don't we just throw it out? Do we actually need it?

seabaylea:
To be clear - that would mean that the connectionCount logic is that superfluous and can be removed (I'm good if that's the case - it gives us something like a 15% perf hit).

djones6:
The spec I referred to was a draft, apparently never made it to the RFC stage, and appears to have expired - though implementations exist (Firefox and Apache, among others).  HTTP 1.1 allows for the presence of Keep-Alive parameters (but doesn't actually define any), so clients are free to ignore them.

The `max` and `timeout` params are more a courtesy to the client, which is still expected to cope with being sent a `Connection: Close`, or with the connection being dropped between requests.  Whether we choose to send these params with the response headers is independent of whether we want the ability to enforce a limit on the number of requests made on a single Keep-Alive connection.

If we do retain this ability, I would like it to be user-configurable, and default to 'unlimited'.

@carlbrown what would be your preference?

carlbrown:
> If we do retain this ability, I would like it to be user-configurable, and default to 'unlimited'.
>
>@carlbrown what would be your preference?

If we don't have a requirement for it, I'd rather omit it.  I'd like this code to be as simple as possible.

djones6:
@carlbrown I removed the Keep-Alive header entirely.  There is no longer any counting of requests per connection,  but the timeout mechanism remains (which is important, as otherwise we would depend on clients to terminate their connections).

PR #42, opened 2017-09-08
Fix port configuration being ignored

seabaylea:
Setting the port for the HTTPServer() was being ignore, both because we weren't passing the port down to the underlying server and socket, and because PoCSocket wasn't carrying out the host to network byte order conversion on Darwin.

### DIFF
--- a/Sources/HTTP/PoCSocket/PoCSocket.swift
+++ b/Sources/HTTP/PoCSocket/PoCSocket.swift
@@ -200,14 +200,14 @@ internal class PoCSocket {
         #if os(Linux)
             var addr = sockaddr_in(
                 sin_family: sa_family_t(AF_INET),
-                sin_port: UInt16(port),
+                sin_port: htons(UInt16(port)),
                 sin_addr: in_addr(s_addr: in_addr_t(0)),
                 sin_zero:(0, 0, 0, 0, 0, 0, 0, 0))
         #else
             var addr = sockaddr_in(
                 sin_len: UInt8(MemoryLayout<sockaddr_in>.stride),
                 sin_family: UInt8(AF_INET),
-                sin_port: UInt16(port),
+                sin_port: (Int(OSHostByteOrder()) != OSLittleEndian ? UInt16(port) : _OSSwapInt16(UInt16(port))),
                 sin_addr: in_addr(s_addr: in_addr_t(0)),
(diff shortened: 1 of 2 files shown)


Sources/HTTP/PoCSocket/PoCSocket.swift
```diff
@@ -200,14 +200,14 @@ internal class PoCSocket {
         #if os(Linux)
             var addr = sockaddr_in(
                 sin_family: sa_family_t(AF_INET),
-                sin_port: UInt16(port),
+                sin_port: htons(UInt16(port)),
```
ianpartridge, line 203: I think you can use `UInt16(bigEndian:)` (and below).

ianpartridge: @weissi is this the same thing as you did in https://github.com/apple/swift-corelibs-foundation/pull/1185 ?

seabaylea: The use of `htons()` on Linux for host to network byte order conversion is pretty standard - and probably makes it clearer what the code is doing here. The non-Linux case is more opaque.

seabaylea: @ianpartridge do you have any major concerns merging as-is? We can always revisit `htons` vs. alternatives later.

ianpartridge: No major concerns, I was suggesting a possible improvement which we can always make in future.

seabaylea:
Ok - merging as-is and raising an issue to look at `htons` vs alternatives.

seabaylea:
Raised #44.

Sources/HTTP/PoCSocket/PoCSocket.swift
weissi: @ianpartridge / @seabaylea `htons()` is the right thing to use here. The only `UInt16(bigEndian:)` is the wrong way around (usually called `ntohs()`). The only reason I used it and not `ntohs()` is that `ntohs` is a C macro that's not imported into Swift.

weissi: you could use `UInt16(port).bigEndian` but `htons()` is to be preferred as it IMHO carries more semantic meaning.

PR #43, opened 2017-09-08
Update README.md and Jazzy annotations

seabaylea:
Update the Jazzy annotations to include HTTPServer and remove the PoCSocket code, and add a proper README.md that provides some basic getting started information.

If/when this is merged I'll update the gh-pages branch with the generated docs.

### DIFF
--- a/Sources/HTTP/HTTPServer.swift
+++ b/Sources/HTTP/HTTPServer.swift
@@ -28,21 +28,26 @@ public protocol HTTPServing : class {
 public class HTTPServer: HTTPServing {
     private let server = PoCSocketSimpleServer()
 
+    /// Create an instance of the server. This needs to be followed with a call to `start()`
     public init() {
     }
 
+    /// Start the HTTP server on the given `port`, using `handler` to process incoming requests
     public func start(port: Int = 0, handler: @escaping HTTPRequestHandler) throws {
         try server.start(handler: handler)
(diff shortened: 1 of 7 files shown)


Sources/HTTP/HTTPServer.swift
```diff
@@ -28,21 +28,26 @@ public protocol HTTPServing : class {
 public class HTTPServer: HTTPServing {
     private let server = PoCSocketSimpleServer()
 
+    /// Create an instance of the server. This needs to be followed with a call to `start()`
     public init() {
     }
 
+    /// Start the HTTP server on the given `port`, using `handler` to process incoming requests
```
adellibovi, line 35: I would add `- Parameter` too for port and handler, what do you think?

adellibovi: I saw that in some other place of the code it's missing, if it's okay I would raise a PR fixing the missing docs for parameters.

adellibovi (changes requested):

seabaylea:
Sure - if the of the change in this PR looks ok then I'll merge and you can add the parameter descriptions (without risking merge conflicts).

adellibovi (approved):
Cool 👍  😄

seabaylea:
Thanks @adellibovi. Merged.

Issue #36, opened 2017-08-26
Write Callback Block on Queue vs on Write


carlbrown:
I'm pulling comment from @helje5 from #28 over into this issue, because that issue is getting cluttered.

> The worse aspect is that this calls the done-cb before the queued write has actually finished! This needs to be passed along the stream, that's the whole point.

My reason for not doing this was two-fold:
1. I was concerned about side-effects from passing the completion blocks into the Socket code, either in the form of memory leaks or in the form of deadlocks.
2. I don't see a case where it matters. The best this code can do is to hold onto the CB until after the socket `write()` completes.  But, as the documentation for `write(2)` explicitly says "A successful return from `write()` does not make any guarantee that data has been committed," holding onto the CB that long doesn't provide any more reliability or information that if it was called where it is now. 

There's no guarantee the connection won't be closed before the bytes that were queued will be written to the socket, but there's no guarantee they're going to make it to the far end of the TCP connection either.  

So why does it matter whether the callback happens after the bytes are queued for `write()` or once `write()` had completed? The guarantee (or lack thereof) is the same either way.  By Occam's Razor then, I think the simpler design is better.

I might well be missing something here.  If so, I'd love to know what it is.

Thanks

helje5:
You don't need a done-CB at all if it doesn't provide 'done' functionality. What purpose does it solve in the current setup, none at all - just drop it then ;-)

The reason we added the CB to Johannes original proposal was to add support for proper back pressure handling. Which is somewhat necessary for async based servers ...

P.S.: When talking about sockets, refer to `man 2 send`, not write. The "does not make any guarantee that data has been committed" refers to filesystem buffers that may not have been flushed to disk. The semantics of a write on a socket file handle are different.

carlbrown:
> The reason we added the CB to Johannes original proposal was to add support for proper back pressure handling. Which is somewhat necessary for async based servers ...

I need to understand better how this would work, because even if I waited to call the CB until after `send()`, I'm don't understand how that would help.  Is there documentation or a book or something you can point me to?

The first indication of congestion we would get (I think) is that the `send()` call would return fewer bytes sent than requested (in which case we spin up a `DispatchSourceWrite` to handle the retry).  In that case, would the CB just be held until the `DispatchSourceWrite` was done? Or should there be a "partially written" value for the callback?  And then what would the application do with that information?

I'm happy to fix this (or review Pull Requests), but I'm not clear on what would need to happen.

[...]

PR #47, opened 2017-09-12
Completion Block moved to socket write Issue#36

carlbrown:
Fix for #36 (see discussion there)


carlbrown:
This needs to be rewritten to honor the stop flag in HTTPBodyHandler - see #48

Closing - will resumbit once that's done.

Issue #48, opened 2017-09-12
Honor stop (Bool) flag in HTTPBodyHandler


carlbrown:
The current implementation doesn't honor the `stop` flag (the `inout Bool`) in `HTTPBodyHandler`. Need to fix that and write a test for it.

anayini:
@carlbrown - Was this issue resolved by #51?

Issue #46, opened 2017-09-12
Alternate transfer encoding (e.g. content-length) needs work and testing


carlbrown:
In working on writing a framework on top of this one, I've discovered that I quite often want to send a payload that has a fixed length (e.g. a `JSON` `Data` produced by `Codable`), and the code paths to do that (e.g. in `HTTPStreamingParser.swift` when `isChunked == false`) are hard to use and poorly tested.

I need to fix that.

PR #49, opened 2017-09-14
Completion block fix

carlbrown:
Fix for completion block issue #36.  

Also includes a fix for #48 that caused me to retract my last attempt (PR #47).

Also adds at least one test for #46 while I was at it (but more tests would still be useful there).


Sources/HTTP/HTTPStreamingParser.swift
```diff
@@ -284,21 +304,29 @@ public class StreamingParser: HTTPResponseWriter {
     func bodyReceived(data: UnsafePointer<Int8>?, length: Int) -> Int32 {
         processCurrentCallback(.bodyReceived)
         guard let data = data else { return 0 }
+        if shouldStopProcessingBody {
```
seabaylea, line 307: Optimally, this should be before the assignment.

Sources/HTTP/HTTPStreamingParser.swift
```diff
@@ -284,21 +304,29 @@ public class StreamingParser: HTTPResponseWriter {
...
-                            finished = true
-                        }), &stop)
+                        //OK, this sucks.  We can't access the value of the `inout` inside this block
+                        //  due to exclusivity. Which means that if we were to pass a local variable, we'd
+                        //  have to put a semaphore or something up here to wait for the block to be done before
+                        //  we could get its value and pass that on to the instance variable. So instead, we're
+                        //  just passing in a pointer to the internal ivar. But that ivar can't be modified in
+                        //  more than one place, so we have to put a semaphore around it to prevent that.
+                        _shouldStopProcessingBodyLock.wait()
+                        handler(.chunk(data: chunk, finishedProcessing: {self._shouldStopProcessingBodyLock.signal()}), &_shouldStopProcessingBody)
```
seabaylea, line 327: `&_shouldStopProcessingBody` isn't protected by the lock....

seabaylea: Do we actually need to wait for the async completion handler to run before we can set stop? Don't we want to set it immediately (sync) even if an async `writeBody()` is still going to happen?

seabaylea:
Out of interest, is there any reason why stop would be a flag rather than an API call on the HTTPResponseWriter, i.e. alongside done() and abort()?

carlbrown:
>Out of interest, is there any reason why stop would be a flag rather than an API call on the HTTPResponseWriter, i.e. alongside done() and abort()?

It's not my API, I just implemented it.

Sources/HTTP/HTTPStreamingParser.swift
carlbrown: Locking the semaphore to check the Bool value should take precedence over the nil-check?

Sources/HTTP/HTTPStreamingParser.swift
carlbrown: > &_shouldStopProcessingBody isn't protected by the lock....

?
That's what line 326 does.

> Do we actually need to wait for the async completion handler to run before we can set stop? Don't we want to set it immediately (sync) even if an async writeBody() is still going to happen

We can't set `stop` here - it's passed as in `inout` to the API, and it's set by the `handler`.

seabaylea: The completion handler that releases the lock runs async (separate thread) - there's no guarantee that its going to call signal() after `_shouldStopProcessingBody` is modified, so its effectively unprotected.

I was wondering whether your intent was to block access to `_shouldStopProcessingBody` until after the completion handler has run.

Sources/HTTP/HTTPStreamingParser.swift
seabaylea: Sorry, I'm getting ahead of myself. I'm not convinced we need any lock protection around a stop flag. We're dealing with two serial queues - one reader and one writer. The only logic that runs on the writer queue is the writeXXXX() calls, so if reads and updates to the stop flag are only by code executing on the reader queue then we don't need a lock.

If we don't need to lock it, it should optimally be before the nil check and assign (unless data = nil is a more common event?)

ianpartridge (commented):
Please break this down into one PR per problem, or a series of commits that build on each other and can be reviewed independently.

carlbrown:
Removing in favor of #51 (and the several others that will be stacked on top of it)

PR #51, opened 2017-09-20
Honor &stopProcessingBody flag, with tests

carlbrown:
Fix for Issue #48, with tests.

Part of what used to be #49

PR #52, opened 2017-09-20
Closes requests with Connection: close header

### LINKED ISSUE #50: Linux: Sockets Not Properly Closed Using `Connection: close` Header
When explicitly declaring the `Connection: close` header on a client request, the server successfully completes the request, but fails to properly close the socket.

Looking deeper into this, it appears due to the below code block in `HTTPStreamingParser` in `Done()`. Here,  `self.parserConnector?.closeWriter()` is called instead of `self.parserConnector?.responseComplete()`, which does not set `self?.responseCompleted = true;` thus, preventing `PoCSocketConnectionListener.close()` from passing the preconditions required to close the socket.

This only occurs on linux, which I think is due to URLSession replacing the explicit `Connection: close` header with a `Connection: keep-alive` header, but I'm not positive about that.

[Code Block](https://github.com/swift-server/http/blob/develop/Sources/HTTP/HTTPStreamingParser.swift#L459-467)

```
public func done(completion: @escaping (Result) -> Void) {
    ...
    let closeAfter = {
            if self.clientRequestedKeepAlive {
                self.keepAliveUntil = Date(timeIntervalSinceNow: 
    StreamingParser.keepAliveTimeout).timeIntervalSinceReferenceDate
                self.parserConnector?.responseComplete()
            } else {
                self.parserConnector?.closeWriter()
            }
        }
    ...
}
```

```
   func close() {
        self.shouldShutdown = true
        
      // Always returns since responseComplete is false and no error has occurred
        if !self.responseCompleted && !self.errorOccurred {
            return
        }
        
        ...

    }
```
Below is a simple example. You'll notice in the terminal `Closing idle socket (#)` is continuously printed until the expectation timeout and the `serverConnection.count` is not decremented
``` 
 func testCloseConnections() {
        let expectation = self.expectation(description: "0 Open Connection")
        let server = HTTPServer()
        do {
            try server.start(port: 0, handler: OkHandler().handle)
            
            let session = URLSession(configuration: URLSessionConfiguration.default)
            let url1 = URL(string: "http://localhost:\(server.port)/echo")!
            var request = URLRequest(url: url1)
            request.httpMethod = "POST"
            request.setValue("close", forHTTPHeaderField: "Connection")
            
            let dataTask1 = session.dataTask(with: request) { (responseBody, rawResponse, error) in
                XCTAssertNil(error, "\(error!.localizedDescription)")
                DispatchQueue.main.asyncAfter(deadline: .now() + 25) {
                    XCTAssertEqual(server.connectionCount, 0)
                    expectation.fulfill()
                }
                
            }
            dataTask1.resume()
            
            self.waitForExpectations(timeout: 30) { (error) in
                if let error = error {
                    XCTFail("\(error)")
                }
            }
            server.stop()
        } catch {
            XCTFail("Error listening on port \(0): \(error). Use server.failed(callback:) to handle")
        }
    }
```

sandmman:
A potential solution to fix #50. The problem occurs on linux because the request is never marked as complete preventing the close() method from executing. 

This creates a blueprint method on the ParserConnecting protocol `func responseCompleteCloseWriter()` that marks a request as complete and closes the socket. 

Also, adds an appropriate test case and fixes the`testMultipleRequestWithoutKeepAliveEchoEndToEnd()` which was broken by the previous merge.


seabaylea:
Hi @sandmman . Merging #51 has meant that there's now a merge conflict. Can you update?

seabaylea:
LGTM. Merging.

PR #53, opened 2017-09-21
Adds Travis Support

sandmman:
Adds support for travis ci

Tests currently fail on Darwin, this is addressed in both #52 #51

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
-
+branches:
+  except:
+  - gh-pages
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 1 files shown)


seabaylea:
Eventually we'll integrate into the Swift CI, but it in the short term Travis makes sense. We'll probably need to scope the branches we test against - primarily to avoid testing against gh-pages (which will fail as there's no source there).

seabaylea:
It looks like this is good to merge, but I'd like to get #52 in first so we don't fail on Darwin.

seabaylea:
#52 is now merged, so adding this in.l

PR #54, opened 2017-09-23
Add public initializer of HTTPRequest

takebayashi:
It is unable to instantiate the struct from external packages unless the public initializer is explicitly  provided.

### DIFF
--- a/Sources/HTTP/HTTPRequest.swift
+++ b/Sources/HTTP/HTTPRequest.swift
@@ -19,6 +19,14 @@ public struct HTTPRequest {
     public var httpVersion: HTTPVersion
     /// HTTP request headers
     public var headers: HTTPHeaders
+
+    /// Create a new HTTP request
+    public init(method: HTTPMethod, target: String, httpVersion: HTTPVersion, headers: HTTPHeaders) {
+        self.method = method
+        self.target = target
+        self.httpVersion = httpVersion
+        self.headers = headers
+    }
 }
 
 /// Method that takes a chunk of request body and is expected to write to the ResponseWriter


seabaylea:
Hi @takebayashi . The lack of public initialiser for `HTTPRequest` is currently deliberate. 

The API design is such that a server receives a HTTPRequest and is provided with a HTTPResponseWriter with which to create a response. The client API (when its added) should be the mirror image of that: using a HTTPRequestWriter to create the request and receiving back a formed HTTPResponse.

How are you looking to use HTTPRequest?

takebayashi:
@seabaylea 
If we try to create a custom implementation of `HTTPServing` we probably need to initialize `HTTPRequest` directly.

Isn't it intended `HTTPServing` to be implemented in custom classes?

carlbrown:
@seabaylea - Last question was to you.

helje5:
@takebayashi No, this is supposed to be a concrete implementation. Not a set of protocols you can plug different backends into (and I suggest to drop remaining protocols, see #85).

However, if you want to do an own custom implementation, you can just create an own API compatible `HTTP` module. See the Apache one for an example: https://github.com/modswift/http/tree/implementation/mod_swift-0.1.0

P.S. A public initialiser would still be useful I think, but for other reasons (e.g. a basic client implementation which we could easily add to this `HTTP` module). But that would be a different Issue I think :-)

seabaylea:
I'm going to close this as yes, the aim is to move towards a single concrete implementation of the API.

takebayashi:
I got it. thank you for answering.

PR #57, opened 2017-09-26
Keep alive override

sandmman:
This exposes the keepAliveTimeout variable to the PoCSimpleServer. This will allow us to create unit tests that are timeout dependent and run in a timely manner.

This also incorporates the change into `testExplicitCloseConnections ` in order to shorten the time taken by test.


carlbrown (approved):
LGTM

PR #58, opened 2017-10-01
TestResponseResolver logic match HTTPStreamingParser

seabaylea:
The logic in TestResponseResolver should match what's used in HTTPStreamingParser. This updates the TestResponseResolver with the changes around `stop` and `finishedProcessing()` handling.

In particular this removes the while loop that has scope to introduce repeated data.


carlbrown (approved):
Makes sense to me

seabaylea:
Thanks @carlbrown. Merging.

PR #62, opened 2017-10-03
Handle failure setting Blocking status of socket more gracefully

carlbrown:
I should not have put a `try!` there in the first place.

### DIFF
--- a/Sources/HTTP/PoCSocket/PoCSocketConnectionListener.swift
+++ b/Sources/HTTP/PoCSocket/PoCSocketConnectionListener.swift
@@ -195,10 +195,23 @@ public class PoCSocketConnectionListener: ParserConnecting {
 
     /// Starts reading from the socket and feeding that data to the parser
     public func process() {
-        try! socket?.setBlocking(mode: true)
-
-        let tempReaderSource = DispatchSource.makeReadSource(fileDescriptor: socket?.socketfd ?? -1,
-                                                             queue: socketReaderQueue)
+        let tempReaderSource: DispatchSourceRead
+        //Make sure we have a socket here.  Don't use guard so that
+        //  we don't encourage strongSocket to be used in the
+        //  event handler, which could cause a leak
+        if let strongSocket = socket {
+            do {
+                try strongSocket.setBlocking(mode: true)
+                tempReaderSource = DispatchSource.makeReadSource(fileDescriptor: strongSocket.socketfd,
+                                                                     queue: socketReaderQueue)
+            } catch {
+                print("Socket cannot be set to Blocking in process(): \(error)")
+                return
+            }
+        } else {
+            print("Socket is nil in process()")
+            return
+        }
 
         tempReaderSource.setEventHandler { [weak self] in
             guard let strongSelf = self else {


Sources/HTTP/PoCSocket/PoCSocketConnectionListener.swift
```diff
@@ -195,10 +195,23 @@ public class PoCSocketConnectionListener: ParserConnecting {
 
     /// Starts reading from the socket and feeding that data to the parser
     public func process() {
-        try! socket?.setBlocking(mode: true)
-
-        let tempReaderSource = DispatchSource.makeReadSource(fileDescriptor: socket?.socketfd ?? -1,
-                                                             queue: socketReaderQueue)
+        let tempReaderSource: DispatchSourceRead
```
ianpartridge, line 198: Not compiled:

```swift
guard let socket = socket else {
    preconditionFailure("Entered \(#function) but socket was nil")
}

do {
    try socket.setBlocking(mode: true)
} catch {
    assertionFailure("Failed to set socket as blocking")
    return
}

let tempReaderSource = DispatchSource.makeReadSource(fileDescriptor: socket.socketfd, 
                                                     queue: socketReaderQueue)

...
```

carlbrown: I specifically put a comment in the next line explaining why I don't think a guard is appropriate.

ianpartridge: Sorry, I don't understand the comment.  How does `guard` encourage anything? It just makes sure the error case is handled clearly.

"Strong" is the wrong terminology for all this anyway, unwrapping an optional does not make it strong in Swift parlance.

Sources/HTTP/PoCSocket/PoCSocketConnectionListener.swift
```diff
@@ -195,10 +195,23 @@ public class PoCSocketConnectionListener: ParserConnecting {
...
+                print("Socket cannot be set to Blocking in process(): \(error)")
+                return
+            }
+        } else {
+            print("Socket is nil in process()")
+            return
+        }
 
         tempReaderSource.setEventHandler { [weak self] in
             guard let strongSelf = self else {
```
ianpartridge, line 217: For example, this line doesn't make sense to me because I don't think it can ever fail. `self` is never nil. Or if it can be, perhaps you could explain under what circumstances that would happen. I don't get it...

seabaylea: The `self` is taken from the closure context (its not passed in as a parameter). As its only weakly referenced its not sufficient to keep it alive, so its possible that `self` no-longer exists.

This is a reasonable design pattern - I'm sure its used elsewhere.

Sources/HTTP/PoCSocket/PoCSocketConnectionListener.swift
seabaylea: I'm not sure I understand this either. Can you help me understand how the use of `guard` vs. `if let` affects the likelihood that someone in the future will make a change to use `strongSocket` in the closure, and indeed why having both `strongSocket` and `socket` helps?

Sources/HTTP/PoCSocket/PoCSocketConnectionListener.swift
ianpartridge: Thanks.  Wouldn't it be safer to set the event handler to `nil` when `self` is destroyed.  It's a class and that's what `deinit{}` is for?

In fact, we already have a `cleanup()` function which does exactly that, and it's called 4 lines below this.

I find it extremely difficult to reason about all this code.

helje5: I didn't follow all this through, but if this is tightly coupled and one can't exist w/o the other, it should not be weak in the first place, but `[unowned]`. Weak is way more expensive for no gain in such a case.
(but be careful w/ threading issues)

seabaylea: Yeah, so it comes down to whether the `readerSource` can (should) exist without its associated `PoCSocketConnectionListener` and `socket`. It looks on the surface like it shouldn't and that the lifecycle of each is tightly coupled. 

My guess is that this code is being conservative and avoiding a cyclic reference should `cleanup()` fail to get called (which in itself should be a bug).

carlbrown: Not only is it possible for `self` to go away while the `readerSource` still exists, but it happens fairly often under load.  See the stack trace of the crash in #69 for just one example.

ianpartridge: OK and that's by design is it?

carlbrown: It's not ideal, but with this API design, I think it's unavoidable.

Sources/HTTP/PoCSocket/PoCSocketConnectionListener.swift
seabaylea: I still don't have a clear understanding if why this approach is needed, i.e.;
* Why guard is inappropriate
* Why the use of `strongSocket` when `socket` is never a weak reference.

Having said that, I'd rather get the fix in and clean this up later if thats the only way to make progress.

PR #63, opened 2017-10-03
Sockets aren't idle if response not completed

carlbrown:
Fixes bug where a long response can be closed if the idle timeout fires while it's in progress

### DIFF
--- a/Sources/HTTP/PoCSocket/PoCSocketConnectionListener.swift
+++ b/Sources/HTTP/PoCSocket/PoCSocketConnectionListener.swift
@@ -151,6 +151,10 @@ public class PoCSocketConnectionListener: ParserConnecting {
 
     /// Check if the socket is idle, and if so, call close()
     func closeIfIdleSocket() {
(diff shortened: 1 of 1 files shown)


ianpartridge:
I'm not qualified to review the queueing aspects of this, for example why `responseCompleted` can be safely modified outside the `socketWriterQueue`.

ianpartridge (commented):
I'm not qualified to review the queueing aspects of this, for example why `responseCompleted` can be safely modified outside the `socketWriterQueue`.

ianpartridge:
Can we add a test for this bug?

carlbrown:
> why `responseCompleted` can be safely modified outside the `socketWriterQueue`.

`responseCompleted` has its own lock, so it's safe to modify, even in the middle of a request/response.  The trick is that we shouldn't act on that in the middle of a read/write, which is why the `close` calls are still queued instead of happening immediately.

seabaylea:
LGTM. Merging.

PR #70, opened 2017-10-07
Use parameter default value to combine init

chrisamanse:
Use Swift default parameters to combine `init(code:reasonPhrase)` and `init(code:)` into one init.

### DIFF
--- a/Sources/HTTP/HTTPResponse.swift
+++ b/Sources/HTTP/HTTPResponse.swift
@@ -89,17 +89,10 @@ public struct HTTPResponseStatus: Equatable, CustomStringConvertible, Expressibl
 
     /// Creates an HTTP response status
     /// - Parameter code: The status code used for the response status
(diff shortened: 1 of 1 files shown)


seabaylea:
Hi @chrisamanse. I'm undecided by this.... I don't think it adds much in the way of code safety (we're already chaining constructors), and effectively retains the two constructors (from an API perspective), but squashes the API doc down which I think makes it harder to read/understand.

Do you feel strongly about getting this in?

chrisamanse:
I agree, this does not add value to code safety. The only added benefit I think is less code, and no duplicate doc comments. But I do understand now that from the API perspective, it would be easier to read, and most users would often use `init(code:)`. In fact, it is rare that users would provide a custom reason. Thus, I do not feel strongly about this anymore.

seabaylea:
Thanks Chris. I'll close this one then.

PR #72, opened 2017-10-07
Use 'guard a == b else' instead of 'if a != b'

chrisamanse:
Improve readability of `HTTPVersion` `Comparable` implementation.


ianpartridge:
I don't really see the benefit. They're both quite readable.

chrisamanse:
It maybe just a personal code style preference. Closing PR since this does not really add any benefits.

PR #73, opened 2017-10-07
Make hashValue a getter instead of a stored value

chrisamanse:
No need for the `hashValue` to be stored. Unless, was this made for performance reasons?

### DIFF
--- a/Sources/HTTP/HTTPHeaders.swift
+++ b/Sources/HTTP/HTTPHeaders.swift
@@ -130,13 +130,11 @@ extension HTTPHeaders {
     public struct Name: Hashable, ExpressibleByStringLiteral, CustomStringConvertible {
         let original: String
         let lowercased: String
-        public let hashValue: Int
 
         /// Create a HTTP header name with the provided String.
         public init(_ name: String) {
             original = name
             lowercased = name.lowercased()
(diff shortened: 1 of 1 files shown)


seabaylea:
If remove the stored property we save on memory, but have to compute the value. `lowercased.hashValue` calls `String.hashValue()`:
https://github.com/apple/swift/blob/master/stdlib/public/core/StringHashable.swift#L133
which in turn calls the internal `_hashString(_ string: String)` function:
https://github.com/apple/swift/blob/master/stdlib/public/core/StringHashable.swift#L87
which has to do a fair amount of work.

In our case, as we store the named header in a Dictionary, the hash performance is relevant - particularly as these are constants that we expect to use/reuse multiple times.

I suspect the approach we have now is the best memory/performance tradeoff, however that's not been tested/verified by comparing the approaches with a relevant benchmark.

chrisamanse:
That's a fair point. In my opinion, it should be the compiler's job to do the optimizations. However, based on the SIL output of `HTTPHeaders.swift` alone, it is not optimized in this step of the compiler. I'm okay with closing this for now until benchmarks of its usage prove that there is no significant difference in performance.

seabaylea:
Thanks @chrisamanse. I'll close this one for now.

PR #75, opened 2017-10-18
server.stop shouldn't have been left commented out in these two tests

carlbrown:
Guessing I was troubleshooting and commenting things out until I found what I was looking for and then failed to uncomment these lines.


seabaylea:
LGTM. Merging.

PR #77, opened 2017-10-23
Fix readBuffer memory leak

chrisamanse:
Use `defer` statement to deallocate `readBuffer` pointer. Currently, it will not be deallocated when `try strongSelf.socket?.socketRead(into: &readBuffer, maxLength:maxLength)` fails, thus causing a memory leak.


seabaylea:
Good catch @chrisamanse. Can you fix the merge conflict? If so, we'll get this in.

chrisamanse:
Fixed merge conflict

seabaylea:
All the macOS builds (for all projects) seem to be failing on Tavis at the moment. I'll merge this once that's fixed and we can get a clean build through.

seabaylea:
We _finally_ got Travis to complete a macOS build. Merging.

PR #78, opened 2017-10-23
TLS integration

gtaban:


gtaban:
Accidentally closed because of change in branch name. I will open another one once #81 is merged.

PR #85, opened 2017-11-05
Drop HTTPServing

helje5:
The API is concrete, this protocol serves no purpose.


seabaylea:
Yeah, I think the original reason was to separate the PoC Server implementation from the HTTP API layer, but as the server implementation doesn't leak out we can change it the future without affecting the API.

Unfortunately there's now a merge conflict - if you don't mind updating we'll get this in.

carlbrown:
@helje5 Will this still be relevant after #96 gets completed?

helje5:
This is distinct to #96, that still has it (I tried not to change too many things). And who knows whether #96 is actually going to be used. Even if not, we should drop `HTTPServing` anyways. (just my opinion, it is certainly not me who decides such stuff ...)

carlbrown:
Resolved the merge conflict - waiting on the CI run

PR #86, opened 2017-11-05
Add optional `queue` to handler

helje5:
For async servers, there has to be a synchronization point
for done callbacks. This is what the `queue` is for.

If the handler doesn't dispatch to other queues but stays
within the calling one, it doesn't need to call the done
callbacks via

    queue.async { done() }

but can do a straight

    done()

But if the handler does perform its work async, it need
to use the given queue to call the done, e.g.:

    func workHard(req: HTTPRequest, res: HTTPResponseWriter, queue: DispatchQueue?) {
        DispatchQueue.background.async {
            // do hard work
            queue.async {
               res.writeHeader(status: .ok)
            }
        }
    }

etc.

We are using `DispatchQueue` here. Another option is to add a
non-optional

  HTTPSynchronizationContext protocol

(adapted by DispatchQueue or a uv or sync variant)

### DIFF
--- a/Sources/HTTP/HTTPCommon.swift
+++ b/Sources/HTTP/HTTPCommon.swift
@@ -6,7 +6,7 @@
 // See http://swift.org/LICENSE.txt for license information
 //
 
-import Foundation
+import Dispatch
 
 /// Typealias for a closure that handles an incoming HTTP request
 /// The following is an example of an echo `HTTPRequestHandler` that returns the request it receives as a response:
--- a/Sources/HTTP/HTTPCommon.swift
+++ b/Sources/HTTP/HTTPCommon.swift
@@ -11,7 +11,7 @@ import Foundation
 /// Typealias for a closure that handles an incoming HTTP request
 /// The following is an example of an echo `HTTPRequestHandler` that returns the request it receives as a response:
 /// ```swift
-///    func echo(request: HTTPRequest, response: HTTPResponseWriter ) -> HTTPBodyProcessing {
+///    func echo(request: HTTPRequest, response: HTTPResponseWriter, queue: DispatchQueue? ) -> HTTPBodyProcessing {
 ///        response.writeHeader(status: .ok)
 ///        return .processBody { (chunk, stop) in
 ///            switch chunk {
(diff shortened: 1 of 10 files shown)


Sources/HTTP/HTTPCommon.swift
```diff
@@ -29,20 +29,26 @@ import Foundation
...
-/// - Parameter res: a writer providing functions to create an HTTP reponse to the request.
-/// - Returns HTTPBodyProcessing: a enum that either discards the request data, or provides a closure to process it.
-public typealias HTTPRequestHandler = (HTTPRequest, HTTPResponseWriter) -> HTTPBodyProcessing
+/// - Parameters:
+///   - req: the incoming HTTP request.
+///   - res: a writer providing functions to create an HTTP reponse to the request.
+///   - queue: optional, if set, dispatch done callbacks back to this queue (if
+///            the handler processing escapes the calling queue)
+/// - Returns: HTTPBodyProcessing: a enum that either discards the request data, or provides a closure to process it.
+public typealias HTTPRequestHandler = (HTTPRequest, HTTPResponseWriter, DispatchQueue?) -> HTTPBodyProcessing
```
seabaylea, line 38: On Linux you'll need to import Dispatch explicitly to have access to DispatchQueue

ianpartridge: Aside: from Swift 4.1 that won't be true, thanks to https://github.com/apple/swift-corelibs-foundation/pull/1206 🙂

helje5: Though importing Foundation is non-sense in the first place, no Foundation is used here ;->

carlbrown:
@helje5 Will this still be relevant after #96 gets completed?

helje5:
Again this depends on whether we want to use #96 or something like it. If not, we should still have this and not restrict the API artificially.

(but yes, if you merge #96, this can be closed too)

carlbrown:
Based on the last comment (and my current assumption/hope that we're going to try to merge #96), I'm going to  close this.  If we decide we can't use #96 for some reason, then we can reopen it.

PR #87, opened 2017-11-07
[HTTPVersion] Refactor HTTPVersion

anayini:
Changes HTTPVersion to an enum to better reflect the
possible values it should have.

### DIFF
--- a/Sources/HTTP/HTTPStreamingParser.swift
+++ b/Sources/HTTP/HTTPStreamingParser.swift
@@ -320,7 +320,7 @@ public class StreamingParser: HTTPResponseWriter {
                         //  just passing in a pointer to the internal ivar. But that ivar can't be modified in
                         //  more than one place, so we have to put a semaphore around it to prevent that.
                         _shouldStopProcessingBodyLock.wait()
-                        handler(.chunk(data: chunk, finishedProcessing: {self._shouldStopProcessingBodyLock.signal()}), &_shouldStopProcessingBody)
+                        handler(.chunk(data: chunk, finishedProcessing: { self._shouldStopProcessingBodyLock.signal() }), &_shouldStopProcessingBody)
                     case .discardBody:
                         break
                 }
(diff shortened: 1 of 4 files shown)


anayini:
Hey all.  Was just checking out this codebase and getting familiar with it and found a few changes that I thought could be nice.  This PR changes `HTTPVersion` from a `struct` to an `enum`, which is a little nicer to use internally.  It also adds a failable initializer that can attempt to create this `enum` from parsed values.

Wondering if you all agree with this?  If so, I would also like input on updating the tests as I think some of them are no longer needed.

helje5 (commented):
Makes no sense to me. The HTTP version is not restricted and can change any day, be it 1.2 or 2.1.
For the common values you could have a constant, which you can compare to. (but i doubt that makes that much sense either)

seabaylea:
Hi @anayini as Helge says, it was a deliberate decision to choose constants over an enum (there was some discussion on this on the mailing list early on too).

The reasoning is that HTTP versioning is open ended, whereas an enum is a closed structure. By using constants we can extend the set of HTTP versions at any time (indeed users can do that through an extension), which can't easily be done with an enum.

anayini:
@seabaylea, @helje5 - I guess my reasoning was that HTTP Version changes are few and far between and a newly introduced version would potentially require software change on this HTTP Server, correct?  Like if version 3.0 comes out, we may need to update our http parsing code.  Since these versions are so rare (many years apart typically) - it seems to make sense to prevent making arbitrary versions in this type as we would add a new case when they are introduced and update the rest of the software to ensure its compatibility.

Also note - that I failed to introduce a case for `v2_0` here which I need to do.

helje5:
It may require a software change in the HTTP module (though not likely for 1.2 or 2.1, just for a major step like 3.x). Rarely it requires a software change in the library consumer.
If you use an enum, you break all code if you update that enum in any way. (Current) enums are not good for anything which could even possibly change in any way since any change breaks source compat. (maybe we get `open enum` in the future, I heard that there is some discussion on that?)

P.S.: And maybe the minimum is that a server could reject a 1.2 or 3.0 and just close the connection. Or do that protocol version status code. For that it still needs to be able to represent the version number.

anayini:
👌 .  Yeah another thought I just had would be that the failable initializer could fall back into the latest supported version.  Like if I pass in `HTTPVersion(major: 1, minor: 3)` it would resolve internally to `HTTPVersion.v1_1`, but if I pass in `HTTPVersion(major: 4, minor: 2)` it would resolve to `nil`

PR #93, opened 2017-11-09
TLS integration

gtaban:
This PR will give support to HTTPS. Tests have been added for both self-signed and CA-signed certs on Linux and macOS.

The TLS code currently sits in:
// protocol definitions and data types
https://github.com/gtaban/security.git 

// implementation using OpenSSL on Linux and SecureTransport on macOS
https://github.com/gtaban/TLSService.git

Once the HTTPS work is integrated, TLS repos can be merged under swift-server.

NB/ Previous PR before I smashed my git branches: https://github.com/swift-server/http/pull/78


carlbrown:
Deferring this until after #96 is completed and merged (and if that doesn't happen, we can reopen this one).

PR #95, opened 2017-11-10
Divide up the Server unit tests and end-to-end tests

gtaban:


carlbrown (approved):
I think this is cleaner.

PR #96, opened 2017-11-10
GCD based, Asynchronous Implementation of the API

helje5:
This PR contains a GCD DispatchIO based implementation of the API and presumably should scale significantly better than the synchronous one. Throughput should be somewhat worse for obvious reasons.

I also setup a small test server demonstrating the API: [http-testserver](https://github.com/ZeeZide/http-testserver). Of special interest is the [`slow` handler](https://github.com/ZeeZide/http-testserver/blob/master/Sources/http-testserver/main.swift#L111), which does the typical `sleep()` async demo.

Unlike Noze.io or Node.js it does not use a single network processing queue, but rather a set of base queues as suggested by Johannes. To support that, the handler API has been enhanced with the `queue` argument. All API calls of a handler need to be dispatched back to that (but the handler itself runs on the same queue, so the trivial cases require no extra work).

This implementation is supposed to properly support back pressure and pipelining.

Caveats:
- tests don't work
- little testing has been done, probably has tons of bugs ;-)


carlbrown (changes requested):
I like this approach, but we at least need to get the tests running first before we merge it.  I can help with that, if you think more than one of us can work on it without stepping all over each other.

helje5:
It would be awesome if you would get the tests working again. I don't plan to work on such any time soon, so feel free to do anything you want!

Adjusting the tests might be challenging, because, well, they need to be asynchronous. For Noze.io I have a helper class which tries to deal with the asynchronicity, not sure whether it might help you or not (or even whether this is the correct way to test that stuff): [NozeIOTestCase.swift](https://github.com/NozeIO/Noze.io/blob/master/Tests/netTests/NozeIOTestCase.swift)

The other hurdle is that I made the `HTTPResponseWriter` concrete (i.e. it is not a protocol anymore). I suppose that makes some mocking hard. We could change it back to a protocol, but I think it was suggested that we should avoid them for performance reasons.

helje5:
BTW: I also still wonder whether we would want to allow multiple implementations, or whether that is just bad for a basic core module. E.g. I can imagine that we might be able to use generics to efficiently do something like Apache MPMs, for example:

```swift
let asyncServer = HTTPServer<DispatchMPM>(port: ...)
```

and

```swift
let syncServer = HTTPServer<ThreadedMPM>(port: ...)
```

Both have their unique advantages and disadvantages. But maybe this would make it too complex, not sure.

carlbrown:
> Adjusting the tests might be challenging, because, well, they need to be asynchronous. For Noze.io I have a helper class which tries to deal with the asynchronicity, not sure whether it might help you or not

Ok. Let me go to work on that, and we'll see how far I get.

Sources/HTTP/HTTPServer.swift
```diff
@@ -1,54 +1,336 @@
-// This source file is part of the Swift.org Server APIs open source project
 //
-// Copyright (c) 2017 Swift Server API project authors
-// Licensed under Apache License v2.0 with Runtime Library Exception
+//  HTTPServer.swift
+//  HTTP
 //
-// See http://swift.org/LICENSE.txt for license information
+//  Created by Helge Hess on 22.10.17.
```
anayini, line 5: We probably want to keep the swift.org license stuff, right?

Sources/HTTP/HTTPServer.swift
```diff
@@ -1,54 +1,336 @@
...
+  
+  private  var listenSource   : DispatchSourceRead?
+  
+  internal let queue =
+                 DispatchQueue(label: "de.zeezide.swift.server.http.server")
+  private  var connections = [ HTTPConnection ]()
+  
+  private  var acceptCount       = 0
+  private  let handlerBaseQueues : [ DispatchQueue ] = {
+    var queues = [ DispatchQueue ]()
```
anayini: nit spacing `var queues = [DispatchQueue]()`

Sources/HTTP/HTTPServer.swift
```diff
@@ -1,54 +1,336 @@
...
+    for i in 0..<ProcessInfo.processInfo.processorCount {
+      let q = DispatchQueue(label: "de.zeezide.swift.server.http.handler\(i)")
```
anayini: Interesting - the idea here is that its most efficient to have a queue per processor?

Sources/HTTP/HTTPServer.swift
```diff
@@ -1,54 +1,336 @@
...
+      queues.append(q)
+      q.async { _ = 46 + 2 } // warm up threads
```
anayini: What exactly happens in Dispatch when you do this?  Is there like a fixed cost on a `DispatchQueue` during the first `async` to it?

Sources/HTTP/HTTPServer.swift
```diff
@@ -1,54 +1,336 @@
...
+  }
+  
+  
+  private(set) public var connectionCount : Int32 = 0
+  
+  
+  private func handleListenEvent(on fd        : Int32,
+                                 handler      : @escaping HTTPRequestHandler,
+                                 localAddress : sockaddr_in)
+  {
```
anayini: In general I prefer:
```.swift
private func handleListenEvent(
  on fd: Int32,
  handler: @escaping HTTPRequestHandler,
  localAddress: sockaddr_in
) {
```

which I think is common in many Swift style guides.

Sources/HTTP/HTTPConnection.swift
```diff
@@ -0,0 +1,620 @@
...
+
+
+// MARK: - Static Strings
+
+@inline(__always)
+fileprivate func headerNameForData(_ data: Data) -> HTTPHeaders.Name? {
+  guard !data.isEmpty else { return nil }
+  // TODO: reuse header values by matching the data via memcmp(), maybe first
+  //       switch on length, compare c0
+  guard let s = String(data: data, encoding: .utf8) else { return nil }
```
anayini, line 565: Can put these both in one `guard` if you like

Sources/HTTP/HTTPConnection.swift
```diff
@@ -0,0 +1,620 @@
...
+///
+/// Since HTTP clients can pipeline calls, a connection can run multiple
+/// handlers at the same time. To maintain proper response ordering,
+/// response-writer objects will be 'corked' until the responses ahead of them
+/// are completely written out.
+///
+internal class HTTPConnection : CustomStringConvertible {
+
+  /// The `HTTPServer` which created the connection. If the connection shuts
+  /// down, it will unregister from that server.
```
anayini: Nit: spacing / color violations

Sources/HTTP/HTTPConnection.swift
```diff
@@ -0,0 +1,620 @@
...
+  /// The array of active writers. There can be many due to HTTP pipelining.
+  /// I.e. a new request can arrive while the previous request is still being
+  /// delivered to the client.
+  /// HTTP/1 requires strict ordering of responses. That is, the first response
+  /// must be fully sent before the next can be sent. To support that, inactive
+  /// writers are 'corked' and will be flushed when they are due.
+  private var writers = [ HTTPResponseWriter ]()
+    // FIXME: make this a linked list
+  
+  internal init(fd             : Int32,
```
anayini, line 60: Nit: usually use newline before first arg when splitting args up on multiple lines

Sources/HTTP/HTTPConnection.swift
```diff
@@ -0,0 +1,620 @@
...
+             "writer done, but it is not the head writer? \(self) \(writer)")
+      return
+    }
+    
+    if let idx = writers.index(where: { $0 === writer }) {
+      // break retain cycle
+      writers.remove(at: idx)
+    }
+    else {
+      assert(false, "did not find writer which went done: \(self)")
```
anayini: Can use `assertionFailure(message: String)`

Sources/HTTP/HTTPServer.swift
helje5: Nah, I wrote it. It should have my name on it 😬 License is fine though.

Sources/HTTP/HTTPServer.swift
helje5: Yeah, that is ugly. But feel free to run one of your style 'cleanups', I don't care.

Sources/HTTP/HTTPServer.swift
helje5: That is not an idea, that is a fact ;-) There is also `activeProcessCount` which might be more appropriate (is that ever different?)

All this really belongs into the Options, care to file an Issue for this? E.g. the above is good for a server, but an on-device server may only want to use one thread.

Sources/HTTP/HTTPServer.swift
helje5: See the thread with Johannes on the mailing list. My takeaway was that GCD only assigns threads on-demand (which can take 100ms on Linux or something I don't know).
But maybe this code is non-sense, a GCD expert should have a look over the whole codebase.

Sources/HTTP/HTTPServer.swift
helje5: I don't, but I don't care much either.

Sources/HTTP/HTTPConnection.swift
helje5: Why would I do this? Those are separate things. This one doesn't make sense for me.

Sources/HTTP/HTTPConnection.swift
helje5: *color violations*??? :-)

2 spaces are right, has ARI approval: http://www.alwaysrightinstitute.com/swifter-space/

but feel free to change it.

Sources/HTTP/HTTPConnection.swift
helje5: Don't care.

Sources/HTTP/HTTPServer.swift
anayini: 👍 Sounds good

Sources/HTTP/HTTPServer.swift
anayini: Hm cool didn't know about that

> Whereas the processorCount property reports the number of advertised processing cores, the activeProcessorCount property reflects the actual number of active processing cores on the system. There are a number of different factors that may cause a core to not be active, including boot arguments, thermal throttling, or a manufacturing defect.

I guess they can be different

helje5:
I enabled Issues on my fork https://github.com/ZeeZide/http/issues and marked those which may be necessary to merge this PR with a `Required for PR` label.

Sources/HTTP/HTTPServer.swift
benasher44: Shouldn't this be the responsibility of `libdispatch`? Having 1 warm thread per processor is great, but GCD is (partly among other things) supposed to free us from having to worry about this detail (# of threads per process, and arguable # of active threads at all).

If we think it'll be commonly desirable to warm threads for this kind of application (a server written in Swift), can we move this logic behind a GCD API or have GCD (if it doesn't do this already) warm threads in a library constructor function? At the very least, I think we should open a ticket/issue and leave a TODO here to come back and clean this up.

benasher44: `DispatchQueue.concurrentPerform` already leans on the internal function `_dispatch_qos_max_parallelism` to understand the capabilities of system in terms of max parallelism, which contains some logic that is like missed here (or at least repeated). I think that if we could figure out a way to get some help from GCD for this purpose, we'd end up with a better (at least in terms of less repeated code, if not in terms of better performance) result.

helje5: > Shouldn't this be the responsibility of libdispatch?

Johannes suggested on the list that this may not be a good idea because the scheduling is application specific and GCD doesn't know about the application needs (I couldn't fully follow his arguments to be honest). My assumption is/was generally the same, it should be GCD which is dealing with it. It just doesn't appear to work well in the real world? No idea :-)

Three examples I can think of:
- on a device you probably just want to run on a single CPU. (though creating the queue as .background may have the same effect)
- on the server you may want to run a specific number of threads if other servers are assigned to different CPUs (i.e. you want to share load and not grab everything available)
- on the server you may want to run more than a single thread per CPU if you have handlers which are not perfectly async

I find the discussion very interesting, but would like to suggest to move it to the mailing list.

FWIW I just noticed that I may have had a flaw in my thinking that you need serial queues to base serial queues upon. Presumably you can also just attach the connection serials queues on a single concurrent target queue just fine.

Sources/HTTP/HTTPConnection.swift
helje5: Because that got a thumb down, I guess I should clarify that what I meant is that I don't care whether or not we do this. I'm happy to do either :-) I don't care whether it is `assert(false)` or `assertionFailure`.

Sources/HTTP/HTTPServer.swift
helje5: I created an issue for this in my fork: https://github.com/ZeeZide/http/issues/12

Sources/HTTP/HTTPConnection.swift
helje5: This actually sounds a little weird, is there a style guide or standard lib example which does it like this?

anayini: Hm - yeah maybe this is just a thing that we do in my company's internal style guide.  

I actually don't even know, is there an Apple approved style guide?  Couldn't find one with a quick search.

helje5: You should find out if you propose such changes ;->

anayini: https://github.com/apple/swift/blob/master/stdlib/public/core/BidirectionalCollection.swift#L166

anayini: I guess they do this only if the line would have been too long without, in which case they start the arguments on the next line.

I can go check to see if other projects are adhering to some style guide somewhere.  Eventually we will probably run into a bunch of different tiny things like this.

helje5: This looks really ugly, but fair enough. On the pro side, it uses a proper 2-space indent :-> Maybe we should have an issue to track what style we want to do?

Just to be clear, I do care a lot about style (and know very well that mine is the only correct one), but I think that in a project like this it should just match the mainline - please never argue with `we do in my company's internal style guide` but with Swift does it like this, check [here].

anayini: 👍 .  Will file an issue.

Sources/HTTP/HTTPServer.swift
helje5: Not that I care a lot but I only know this style guide: https://github.com/raywenderlich/swift-style-guide and that doesn't have it ...

Sources/HTTP/HTTPConnection.swift
anayini: Wasn't really arguing with `we do this in my company's internal style guide`.  Only that I'm used to that style guide so I made the mistake of marking this as wrong here.

carlbrown:
@helje5  I added a [PR](https://github.com/ZeeZide/http/pull/13) against your fork that has a batch of changes for getting tests to run on the Mac. (There will be another one once I find/fix a bug on Linux).

Sources/HTTP/HTTPServer.swift
ogres: This is probably not a good idea.
None of the other files in this repo have neither this nor "Copyright © 2017 SomeCompany..."
Do you think each time someone changes this file, they should add "Changed by..." to it as well? We have a git history for it, so your work won't be forgotten :)

Sources/HTTP/HTTPServer.swift
weissi: just to comment here: In a well behave system I'd say we can expect to be IO bound here, therefore more threads won't help us to be faster. Also there is a latency problem in GCD on Linux when spawning threads because of this: https://github.com/apple/swift-corelibs-libdispatch/blob/666df60ce98b9f38814b544f9ff1ca74b7bbf70d/src/queue.c#L5669 . This however only seems to affect _concurrent_ queues (at least these days).

I wrote a new latency test to assess the situation on Swift 4 with _serial_ queues (see below) and the results are that on Swift 4 with _serial queues_ there are no longer any latency issues. In other words: we can change this code to just use one serial queue per request that isn't targeted at a base queue and it should still work fine.

Without the base queues, I'd expect performance under heavy IO load to be worse as we'd end up with too many threads. There is however a (potentially big) benefit: If a client of this API decides to block (the kernel thread) in a request handler it would no longer an issue. 

I'm happy to switch to one serial queue per request as that would make the code simpler and if someone wants the very best performance it'd be an easy thing to bring back.

```
import Dispatch
import Glibc

let g = DispatchGroup()
g.enter()

func foo(remaining: Int) {
    let q = DispatchQueue(label: "some q")
    if remaining > 0 {
        q.async {
            print("\(pthread_self())")
            foo(remaining: remaining - 1)
        }
        usleep(10_000)
    } else {
        g.leave()
    }
}

foo(remaining: 10000)
g.wait()
```

Sources/HTTP/HTTPServer.swift
weissi: I did indeed say that and it is still the case (on Linux) for _concurrent queues_ but not serial queues. Apologies.

Sources/HTTP/HTTPServer.swift
helje5: > more threads won't help us to be faster

I think keeping the base queues is worth it for the reverse. I also do on-device servers and I would usually prefer to run those on less threads than the device has cores (one or maybe three on the iPhone Y). The same may be true on servers which are not dedicated to just one app.

What do you think?

helje5: Also ```usleep(10_000)``` vs ```foo(remaining: 10000)``` 🤔

weissi: @helje5 the `usleep` is just a crappy way to force it to not re-use the very same thread.

If we want to over-engineer it we could also just have a `baseQueueProvider` closure or something that either returns an appropriate base queue or if the user doesn't want that just `nil`. Because base queue `nil` means no target queue AFAIK.

helje5: Man, this was 10_000 vs 10000 ;-)

The closure doesn't sound too bad. Would be optional and part of the Options object passed in.

helje5: I added the closure and dropped the specific algorithm.

GeorgeLyon:
Don't want to derail this PR, but I think leaking the DispatchQueue the server uses into the handler is mistake. Read and write operations on a descriptor don't need to be synchronized as far as I am aware. An example where the server and handler synchronization primitives are separate can be found here: https://github.com/GeorgeLyon/Server

helje5:
> Don't want to derail this PR, but I think leaking the DispatchQueue the server uses into the handler is mistake.

It is an optimisation. I don't really like it either, but the alternatives seem too expensive to me (require the response-writer to be free-threaded).

> Read and write operations on a descriptor don't need to be synchronized as far as I am aware

Yeah, but I don't see how that helps us in any way. The goal here is to allow for asynchronous handlers which support back pressure. That implies more book-keeping. Which needs to be synchronised.

P.S.: Please lets continue the discussion on-list.

helje5:
Superfluous, just use: https://github.com/apple/swift-nio/tree/master/Sources/NIOHTTP1

PR #101, opened 2017-11-12
Any informational header (1XX) should be allowed to be written prior …

GeorgeLyon:
…to actual response headers.

Previously, this honor was only bestowed upon status 100 "Continue".

### DIFF
--- a/Sources/HTTP/HTTPStreamingParser.swift
+++ b/Sources/HTTP/HTTPStreamingParser.swift
@@ -366,10 +366,10 @@ public class StreamingParser: HTTPResponseWriter {
 
         var header = "HTTP/1.1 \(status.code) \(status.reasonPhrase)\r\n"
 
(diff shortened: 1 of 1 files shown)


helje5:
Sounds good, do you want to do 101 or something else?

GeorgeLyon:
101 is special (and would probably need to be handled separately).
I'm actually hoping to use this to *avoid* using 101 (to upgrade to websockets) and instead use 102 or an application-specific code to enable long-polling.
There is also 103 (currently in draft) which allows you to specify the "Link" header any number of times before the full response is provided.

helje5:
Yes, upgrade has special support in the parser too. Needs to be implemented, right now we don't even have API for that. (in a way we really need a stream abstraction to make this work)

carlbrown (approved):
Looks reasonable to me.

Issue #83, opened 2017-11-03
Proposal: Some suggestions for better, more readble API


amosavian:
I don't know where I can submit proposals for api changes so I opened this issue.

### HTTPServer port

server port is determined in `start(port:, handler:)` method, while I think a server can start once with a one specified port. I could find a `port` property in `HTTPServer` class while with this design it have to be `ports` array. I propose relocating `port` argument from `start()` to class `init()` method.

### HTTPResponse write methods

Personally I prefer something like `response.header.write()` instead of `response.writeHeader()`, `response.body.write()` instead of `response.writeBody()` and `response.trailer.write()` instead of `response.writeTrailer()`. Also we can keep room to extend `header`, `body` and `trailer` properties later.

### HTTP version static variables

HTTP versions are limited, thus having `HTTPVersion.v1_1` and `HTTPVersion.v2_0` static vars might be fair.

### A more complicated `HTTPHeaders`

HTTP headers are simply defined as a pair of strings. But for many of headers there have a set of defined values or a structured value. This implementation is prone to typo and human errors. So I propose to have some setters for frequent ones. 

##### Implementation detail
```swift
// To set 'Range'
let range = 100... // half open or closed
headers.set(range: range) // result: 'Range: bytes 100-'

// To set 'Accept-Encoding'
headers.set(acceptEncodings: [.deflate, .gzip, .brotli, .identity])
// or
headers.set(acceptEncoding: .deflate, quality: 1.0)
headers.add(acceptEncoding: .gzip, quality: 0.8) // Note add instead of set
headers.add(acceptEncoding: .identity, quality: 0.5)
// values are defined in a enum/struct

// To set 'Cookie'
let cookie = HTTPCookie(properties: [.name : "name", .expires: Date(timeIntervalSinceNow: 3600)])
headers.set(cookie: cookie)
...
headers.add(cookie: cookie2)

// To set 'Accept-Language'
let locale = Locale.current
headers.set(acceptLanguages: [locale])
// also 'add(acceptLanguage:)' method to add progressively like Accept-Encoding

// To set 'Accept-Charset'
headers.set(acceptCharset: String.Encoding.utf8)
``` 

For Content-Type, we would define this struct in `HTTPHeaders` (should be extended with more types):
```swift
extension HTTPHeaders {
    struct MIMEType: RawRepresentable {
        public var rawValue: String
        public typealias RawValue = String
        
        public init(rawValue: String) {
            self.rawValue = rawValue
        }
        
[...]

PR #104, opened 2017-11-16
Implemented #83 (HTTPHeaders accessors)

amosavian:
This pull request includes static variables for `HTTPVersion` struct. Also implemented easy accessors for `HTTPHeaders` to strict types for header values.
~`Authentication`, `Cookie` and `Set-Cookie` accessors are not implemented yet due to complexity or designing limitations. I will make another PR when they became ready in near future.~

**UPDATE:** All methods are implemented now. Here is mailing list post: mailing list: https://lists.swift.org/pipermail/swift-server-dev/Week-of-Mon-20171113/000713.html


Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,795 @@
...
+//
+// See http://swift.org/LICENSE.txt for license information
+//
+
+import Foundation
+
+extension HTTPHeaders {
+    
+    /// MIME type directive for `Accept` and `Content-Type` headers
+    public struct ContentType: RawRepresentable, Hashable, Equatable {
```
helje5: This is weird because it can't carry arguments. Maybe this should be called MediaType? And then a ContentType for a content-type?

E.g. how do you do:

    Content-Type: multipart/form-data; boundary=something

or

    Content-Type: text/calendar; component=VCARD

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,795 @@
...
+            case .digest: return "Digest"
+            case .oAuth1: return "OAuth"
+            case .oAuth2: return "Bearer"
+            case .custom(let type): return type
+            }
+        }
+    }
+    
+    /// Encoding of body
+    public enum Encoding: String {
```
helje5: This needs to be open, an `enum` is not appropriate here for obvious reasons. (at the minimum add a `.other`, but that is not really neat either - rather do constants)

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,795 @@
...
+                params[key] = value
+            }
+        }
+        return params
+    }
+    
+    // Request Headers
+    
+    /// Fetch `Accept` header values, sorted by `q` parameter
+    public var accept: [ContentType] {
```
helje5: What about wildcard support? Quality arguments? Supporting this would be the requirement for a structured header:

    Accept: text/html, application/xhtml+xml, application/xml;q=0.9, */*;q=0.8

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,795 @@
...
+            self.storage[.acceptEncoding]!.append(acceptEncoding.rawValue)
+        }
+    }
+    
+    // TODO: Implement var authentication
+    
+    // TODO: Implement var cookie: [HTTPCookie]
+    
+    /// `If-Match` header etag value
+    public var ifMatch: String? {
```
helje5: If-Match takes an array of etags, supports weak and wildcards. For a typed header this seems very insufficient.

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,795 @@
...
+        get {
+            return self.storage[.contentLanguage]?.first.flatMap(Locale.init(identifier:))
+        }
+        set {
+            self.storage[.contentLanguage] = newValue.flatMap { [$0.identifier.replacingOccurrences(of: "_", with: "-")] }
+        }
+    }
+    
+    /// `Content-Length` header value
+    public var contentLength: Int64? {
```
helje5: I'm not convinced `Int64` is right here as you can't expect a x32 implementation to deal w/ x64 Integers. I think that should be rejected at the HTTP level with a proper response.

Sources/HTTP/HTTPVersion.swift
```diff
@@ -20,6 +20,17 @@ public struct HTTPVersion {
     }
 }
 
+extension HTTPVersion {
+    /// HTTP version 1.0
+    public static let v1_0 = HTTPVersion(major: 1, minor: 0)
```
helje5, line 25: I don't know, maybe, why not. I'm not sure when I would ever use those constants.

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,795 @@
...
+        
+        return nil
+    }
+    
+    /// Formats date according to RFCs standard.
+    func format(with standard: RFCStandards, locale: Locale? = nil, timeZone: TimeZone? = nil) -> String {
+        let fm = DateFormatter()
+        fm.dateFormat = standard.rawValue
+        fm.timeZone = timeZone ?? TimeZone(identifier: "UTC")
+        fm.locale = locale ?? Locale(identifier: "en_US_POSIX")
```
helje5: The timezone and locale should at least get cached.

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,795 @@
...
+        public static let usenet = RFCStandards.rfc850
+        
+        // Sorted by commonness
+        fileprivate static let allValues: [RFCStandards] = [.rfc1123, .rfc850, .iso8601, .asctime]
+    }
+    
+    /// Checks date string against various RFC standards and returns `Date`.
+    init?(rfcString: String) {
+        let dateFor: DateFormatter = DateFormatter()
+        dateFor.locale = Locale(identifier: "en_US")
```
helje5: cache

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,795 @@
...
+        get {
+            return self.storage[.date]?.first.flatMap(Date.init(rfcString:))
+        }
+        set {
+            self.storage[.date] = newValue.flatMap { [$0.format(with: .rfc1123)] }
+        }
+    }
+    
+    /// `ETag` header value
+    public var eTag: String? {
```
helje5: the etag header contains an array. also, no support for w/

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,795 @@
...
+        get {
+            return self.storage[.eTag]?.first
+        }
+        set {
+            if let newValue = newValue {
+                // eTag value must be a quoted string, we check only suffix to be compatible with weak eTags
+                let finalValue = newValue.hasSuffix("\"") ? newValue : "\"\(newValue)\""
```
helje5: escaping?

helje5 (commented):
While I'm kinda think that we should have typed headers, I think this PR has tons of issues and doesn't reflect the RFCs.

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: - This property is a getter. setter method(`set(accept:, quality:)`) take quality as a argument, and getter method sorts values. I believe a server implementation doesn't need q value when It's sorted already.
- There is a `ContentType.all` which is wildcard, there are also wildcards for audio, video and text.

helje5: This is intended to be not just for the server, just fwiw.

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: It returns eTags with quotations. so you can have `"W/\"12345\""` as an etag. Maybe we can present value as an associated enum? (e.g. `.strong("12345")`, `.weak("12345")` and `.wildcard("*")`)
We can change design.

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: Apple's `URLResponse.expectedContentLength` is `Int64`, I decided to use Int64 for compatibility. A file can be bigger than 2TB btw.

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: ditto from `If-Match`

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: I implemented `contentCharset` property for `charset` parameter. We can change design btw

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: Encoding should be supported by server. I believe we can't have custom values when they are not supported unless you show me I'm wrong.

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: No, that adds a quotation to newValue if it doesn't have already. for weak etags

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: Yes, I implemented getter and setter method for all headers. They're maybe redundant in some cases.

helje5:
May I also suggest that you take this to the mailing list for discussion. I think we already discussed this kind of stuff and decided not to do typed headers at this level of the framework. It would be part of a real framework on top of this driver.

If we really go the route of typed headers (which I personally kinda like for efficiency), we should actually do that in the storage itself, not just wrapping the String key/value. I wrote some email on that in the list.

amosavian:
@helje5 please check new implementation of EntryTag.

amosavian:
I don't know how to use mailing list unfortunately. I wonder if you could help me.

Headers are not something that you would access frequently. For requests, you may access them once or twice. For response, you set them once. Parsing all request headers when initializing is not an optimized behavior unless they got accessed (lazy). We may introduce some kind of caching later btw.

helje5:
https://lists.swift.org/mailman/listinfo/swift-server-dev

amosavian:
@helje5 I sent email and it's awaiting moderator approval

helje5:
You first need to subscribe to the list with the email address you want to use for sending messages to it. Once this is through, you don't need to wait for moderator approval.

amosavian:
Posted in mailing list: https://lists.swift.org/pipermail/swift-server-dev/Week-of-Mon-20171113/000713.html

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,1229 @@
...
+        /// Chunked body in `Transfer-Encoding` response header
+        case chunked
+        /// Can have trailers in `TE` request header
+        case trailers
+    }
+    
+    /// Determines server accepts `Range` header or not
+    public enum RangeType: String {
+        /// Can't accept Range
+        case none
```
nacho4d: I recently read RFC7233 for implementation somewhere else. RCF allows to define any kind of range not only `bytes`. In the specific cases of File Serving `bytes` just happens to be the most common case. Ranges could be used in other REST, etc calls too and it could be `items`, `rows`, etc whatever, restricting this to only bytes might not allow http module users to use this function to other purposes than file serving.

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,1229 @@
...
+            } else {
+                self.storage[.ifNoneMatch] = nil
+            }
+        }
+    }
+    
+    /// `If-Range` header etag value
+    public var ifRange: EntryTag? {
+        get {
+            return self.storage[.ifRange]?.first.flatMap(EntryTag.init)
```
nacho4d: As of RFC7233 and RFC7232 `ifRange` could also be of the form of `Last-Modified` header. instantiating from EntryTag will not allow the later form.

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: Please find fixed RangeType and ifRange implementations

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,1596 @@
...
+            return self.storage[.origin]?.first.flatMap(URL.init(string:))
+        }
+        set {
+            self.storage[.origin] = newValue.flatMap { [$0.absoluteString] }
+        }
+    }
+    
+    /// Returns `Range` header value.
+    /// - Note: upperbound will be Int64.max in case of open ended Range.
+    public var range: Range<Int64>? {
```
nacho4d: According to RFC7233 range headers can have multiple ranges like: `Range: bytes=3-5,10-15`. So this should be defined as `[Range<Int64>]?`.

Also, the spec for parsing ranges is not as simple as below 4 lines but I guess that can be addressed later as long as the interface is ok :p

amosavian: As far as I understood from examples in RFC, multiple ranges are continuous if we intersect values. e.g:

`bytes=0-0,-1` becomes: -1..<1
`bytes=500-600,601-999` becomes: 500..<1000
`bytes=500-700,601-999` becomes: 500..<1000

Thus it is not an array of range. Please correct me if I'm wrong?

Update: RFC explicitly says ` byte-range-set  = 1#( byte-range-spec / suffix-byte-range-spec )` which means one value is acceptable, and calls multiple ranges as "valid (but not canonical)". All examples are continuous though.

nacho4d: Spec defines various ranges, they do not have to be continuous, for example response in page 11 shows at least two ranges bytes=500-999,7000-7999

As per spec, server CAN merge overlapping or adjacent ranges however that is not a must. Multipart response for partial content is clearly defined which indicates the existence of multiple ranges. 

     HTTP/1.1 206 Partial Content
     Date: Wed, 15 Nov 1995 06:25:24 GMT
     Last-Modified: Wed, 15 Nov 1995 04:58:08 GMT
     Content-Length: 1741
     Content-Type: multipart/byteranges; boundary=THIS_STRING_SEPARATES

     --THIS_STRING_SEPARATES
     Content-Type: application/pdf
     Content-Range: bytes 500-999/8000

     ...the first range...
     --THIS_STRING_SEPARATES
     Content-Type: application/pdf
     Content-Range: bytes 7000-7999/8000

     ...the second range
     --THIS_STRING_SEPARATES--


https://tools.ietf.org/html/rfc7233

nacho4d: Spec defines various ranges, they do not have to be continuous, for example response in page 11 shows at least two ranges bytes=500-999,7000-7999

As per spec, server CAN merge overlapping or adjacent ranges however that is not a must. Multipart response for partial content is clearly defined which indicates the existence of multiple ranges. 

     HTTP/1.1 206 Partial Content
     Date: Wed, 15 Nov 1995 06:25:24 GMT
     Last-Modified: Wed, 15 Nov 1995 04:58:08 GMT
     Content-Length: 1741
     Content-Type: multipart/byteranges; boundary=THIS_STRING_SEPARATES

     --THIS_STRING_SEPARATES
     Content-Type: application/pdf
     Content-Range: bytes 500-999/8000

     ...the first range...
     --THIS_STRING_SEPARATES
     Content-Type: application/pdf
     Content-Range: bytes 7000-7999/8000

     ...the second range
     --THIS_STRING_SEPARATES--


https://tools.ietf.org/html/rfc7233

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,1677 @@
...
+            ranges.append((lower, upper))
+        }
+        return ranges
+    }
+    
+    /// Returns `Range` header values. An empty array means no value is set in header.
+    /// - Note: upperbound will be Int64.max in case of open ended Range.
+    /// - Note: Server response should be `multipart/byteranges` in case of more than one range is returned.
+    ///    See [MDN's HTTP range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) for more info.
+    /// - Important: A negative range means server must return last bytes/items of file.
```
nacho4d: `bytes` is the most common type and to validate ranges one would get the file size from a file as: `(fileAttributes[FileAttributeKey.size] as? NSNumber)?.uint64Value` because file size are declared as `UInt64` in Foundation. However current declaration is `Int64` which means there would be cases when the string can not be converted to Range<Int64>. How can we address this?. 

Example: try to get a tiny part of a very big file: 
`Range: bytes=18446744073709551600-18446744073709551601`

Are you using `Int64` is because of ranges like: `Range: bytes=-3` which look like negative?. Actually there is no such a thing as negative ranges. In fact they should be converted into a  range object/struct **after** the file size is known, for example if fileSize is `1000` then it would mean: `Range: bytes=997-999` which uses only positive numbers.

I like the idea of having everything statically defined as structs/classes/enums rather than raw strings, but I am not sure having this interface will help to correctly implement the RFC :(

(As a side note, nodejs implementation converts range header string into range objects in a function that receives the fileSize as a parameter.)

amosavian: About Foundation: `HTTPURLResponse.expectedContentLength` and `URLSession` methods return `Int64`.

On negative values, We don't define how to interpret data as we don't know about resource. It's upon higher level framework to convert negative values to meaningful ones or return 416.

About BigInt, Using Decimal will hit performance. We are implementing these method for common cases. A file with that size is not common and higher level framework would have to implement theirs. Should we add limitations to documentation?

nacho4d: OK, if other values are Int64 then I think this should be ok :]
IMO I don't think documentation needs to know about this. Other framework implementers will likely read comments in code

amosavian:
@seabaylea @carlbrown Any chance to get merged? I must begin working on tests.

Sources/HTTP/HTTPVersion.swift
dschwartz783: Maybe someone has an old copy of Netscape lying around, and can't use HTTP 1.1. :(

aleksey-mashanov:
Great work!
Typed headers are definitely the thing a good framework should have. But I have a sense that this too much for an API between an HTTP framework and an HTTP server. Maybe this should be made as a separate module?

I suggest to replace `Int64` with `Int` to follow Swift guidelines:
> Unless you need to work with a specific size of integer, always use `Int` for integer values in your code.

amosavian:
@aleksey-mashanov Please find my response for `Int64` [here](https://github.com/swift-server/http/pull/104#discussion_r152649403).  HTTPURLResponse.expectedContentLength and URLSession methods return `Int64`. NSFileManager returns `UInt64` for file size but `Content-Length` can have negative values. Furthermore it's common to have a file larger than 2GB on a 32-bit system.

Please find my explanation [here](https://gist.github.com/amosavian/084636d1a382717b7d7ce7469aaacc07) on why we should implement typed header in this layer. RFC 5987 is a reason e.g.

aleksey-mashanov:
>  Furthermore it's common to have a file larger than 2GB on a 32-bit system.

Agree. But 32bit server? We still have some in production for historical reasons but it is a pain. In almost every place in a surrounding code this value will be converted to `size_t` (`Int`) and used for generation of dynamic content. I really don't know what is better: precise API or comfortable API. For low-level of this HTTP API maybe first but this PR is out of scope of low-levelness.

> RFC 5987 is a reason e.g.

Agree. For me RFC5987 and parsing of additional parameters (after `;`) are the most interesting parts of this PR. Maybe they should be implemented lower in `HTTPHeaders` as an additional subscripts?
There are a lot of RFCs, a lot of additional headers and even more will arise. It will be hard to follow all of them correctly and most of changes will be source compatibility breaking.

Again. This PR doesn't provide features which are required to efficiently pass data from a framework to a server and back. Nevertheless it is very useful.

amosavian:
> every place in a surrounding code this value will be converted to size_t (Int) 

Swift Foundation's `FileManager` class works with file size as `UInt64`. In case we don't want typecast, it's better to consider `UInt64` rather than `Int`. On the other hand, Swift compiler may give a warning if user tries to cast `UInt64` to `Int64` or `Int`.

Furthermore, `range`, `contentRange` and `contentLength` must have same type imo. `range` must support negative values.

> There are a lot of RFCs, a lot of additional headers and even more will arise.

I mentioned RFC5987 just as an example because it's mentioned in source code as TODO. Definitely There are more cases which I'm not aware of.

>  PR doesn't provide features which are required to efficiently pass data from a framework to a server and back. Nevertheless it is very useful.

This design is implemented as an extension to current design. My primary goal was not to touch current architecture because any change in architecture usually needs extensive workgroup paperworks (PS: I'm not aware how swift-server workgroup works). But my design, which uses initializers for casting instead of functions (except parsing parameters and `q` values), gives the flexibility to change underlying design anytime.

amosavian:
List of optional but unimplemented headers, which are nice to have in portfolio. I won't implement them as a part of this PR:
`Access-Control-*`, `Accept-Patch`, `Alt-Svc`, `Content-Security-Policy`, `DNT`, `Forwarded`, `Proxy-Authenticate`, `Proxy-Authorization`, `Retry-After`, `Tk`, `Warning`, `X-XSS-Protection`

amosavian:
This PR is completely implemented.

tanner0101 (approved):
awesome

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,800 @@
...
+    
+    // `Cookie` header value. An empty array means no value is set in header.
+    public var cookieDictionary: [String: String] {
+        let pairs = (self.storage[.cookie]?.joined(separator: ";")).flatMap({
+            HTTPHeaders.parseParams($0, separator: ";", removeQuotation: true)
+        }) ?? [:]
+        return pairs
+    }
+    
+    /// `Origin` header value.
```
amosavian: Change to `host`

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,800 @@
...
+            // TODO: Tolerate base64 string not padded with =, Should we?
+            return self.storage[.contentMD5]?.first.flatMap { Data(base64Encoded: $0) }
+        }
+        set {
+            self.storage[.contentMD5] = newValue.flatMap { [$0.base64EncodedString()] }
+        }
+    }
+    
+    fileprivate func dissectContentRange(_ value: String?) -> (from: UInt64, to: UInt64?, total: UInt64?)? {
+        // Converting real negatives to _ to avoid conflict with - as separator
```
amosavian: Remove this comment

Sources/HTTP/HTTPHeadersAccessors.swift
amosavian: Change comment to `host`

Sources/HTTP/HTTPHeadersTypes.swift
```diff
@@ -0,0 +1,1529 @@
...
+                    self = .custom(type, token: nil, params: HTTPHeaders.parseParams(q))
+                }
+            }
+        }
+        
+        public var rawValue: String {
+            switch self {
+            case .basic(let user, let password):
+                let text = "\(user):\(password)"
+                let b64 = (text.data(using: .ascii) ?? text.data(using: .utf8))?.base64EncodedString() ?? ""
```
amosavian: utf8 is default encoding, change the order

Sources/HTTP/HTTPHeadersTypes.swift
```diff
@@ -0,0 +1,1529 @@
...
+        }
+        
+        /// :nodoc:
+        public subscript(parameterName: String) -> String? {
+            get {
+                return self.parameters[parameterName]?.trimmingCharacters(in: .quoted)
+            }
+        }
+        
+        /// Inits a noew
```
amosavian: Fix comment

Sources/HTTP/HTTPHeadersTypes.swift
```diff
@@ -0,0 +1,1529 @@
...
+        static public let xhtml = MediaType(generalType: "application", type: "xhtml+xml")
+        /// Javascript code file
+        static public let javascript = MediaType(generalType: "application", type: "javascript")
+        /// Javascript notation
+        static public let json = MediaType(generalType: "application", type: "json")
+        
+        // Images
+        
+        /// All Image types
+        static public let image = MediaType(generalType: "image", type: "bmp")
```
amosavian: change type: to `”*”`

Sources/HTTP/HTTPHeadersTypes.swift
```diff
@@ -0,0 +1,1529 @@
...
+            // Value is strong
+            let linted = rawValue.trimmingCharacters(in: .quotedWhitespace)
+            self = .strong(linted)
+        }
+        
+        public var rawValue: String {
+            switch self {
+            case .strong(let etag):
+                // TODO: Remove non ascii characters
+                let lintedEtag = etag.trimmingCharacters(in: .quotedWhitespace)
```
amosavian: remove redundant trimming

Sources/HTTP/HTTPHeadersTypes.swift
```diff
@@ -0,0 +1,1529 @@
...
+                return "\"\(lintedEtag)\""
+            case .weak(let etag):
+                let lintedEtag = etag.replacingOccurrences(of: "W/\"", with: "", options: [.anchored, .caseInsensitive]).trimmingCharacters(in: .quotedWhitespace)
```
amosavian: ditto

Sources/HTTP/HTTPVersion.swift
amosavian: HTTP 1.0 and 1.1 are different only in default values for `Keep-Alive` and `Connection` headers afaik. We will replace instances in current code like [here](https://github.com/swift-server/http/blob/31d066a0e92b4575a4389d592e022fe0b9933956/Sources/HTTP/HTTPStreamingParser.swift#L401)

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,800 @@
...
+    
+    /// `Trailer` header value. An empty array means no value is set in header.
+    public var trailer: [HTTPHeaders.Name] {
+        get {
+            return self.storage[.trailer]?.flatMap({ HTTPHeaders.Name($0) }) ?? []
+        }
+        set {
+            if !newValue.isEmpty {
+                // TOCHECK: Forbidden headers ought to be ignored/dropped
+                self.storage[.trailer] = newValue.map { $0.original }
```
amosavian: replace with `.method`

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,800 @@
...
+    
+    /// `Vary` header value. An empty array means no value is set in header.
+    public var vary: [HTTPHeaders.Name] {
+        get {
+            return self.storage[.vary]?.flatMap({ HTTPHeaders.Name($0) }) ?? []
+        }
+        set {
+            if !newValue.isEmpty {
+                // TOCHECK: Forbidden headers ought to be ignored/dropped
+                self.storage[.vary] = newValue.map { $0.original }
```
amosavian: ditto

Sources/HTTP/HTTPHeadersTypes.swift
```diff
@@ -0,0 +1,1529 @@
...
+        static public let bmp = MediaType(generalType: "image", type: "bmp")
+        /// Graphics Interchange Format photo
+        static public let gif = MediaType(generalType: "image", type: "gif")
+        /// JPEG photo
+        static public let jpeg = MediaType(generalType: "image", type: "jpeg")
+        /// Portable network graphics
+        static public let png = MediaType(generalType: "image", type: "png")
+        /// Scalable vector graphics
+        static public let svg = MediaType(generalType: "image", type: "svg+xml")
+        /// Scalable vector graphics
```
amosavian: Fix comment

Sources/HTTP/HTTPHeadersAccessors.swift
```diff
@@ -0,0 +1,800 @@
...
+    public var accept: [MediaType] {
+        get {
+            return self.storage[.accept].flatMap({
+                HTTPHeaders.parseQuilified($0, MediaType.init(rawValue:))
+            }) ?? []
+        }
+    }
+    
+    /// Fetch `Accept-Charset` header values, sorted by `q` parameter. An empty array means no value is set in header.
+    public var acceptCharset: [String.Encoding] {
```
amosavian: Handling wildcard

mohsen1:
FYI DOM [`Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers) API allows setting multiple values for the same cookie and ignores the key casing:

```js
var h = new Headers()
h.append('Cookie', 'foo bar')
h.append('cooKie', 'baz')
for (var e of h.entries()) { console.log(e) } 
```

prints:

```
["cookie", "foo bar, baz"]
```

ianpartridge:
Closing as this project is historical, per the README.

